PR #75, opened 2024-03-24
Mention how the progress bar updates

carmocca:
## What does this PR do?

The progress bar updates when a worker finishes, not when a worker yields. If you don't know this and your workers take a long time to finish, it looks like it froze

### DIFF
--- a/src/litdata/processing/data_processor.py
+++ b/src/litdata/processing/data_processor.py
@@ -931,7 +931,10 @@ class DataProcessor:
 
         self._cleanup_cache()
 
-        print(f"Starting {self.num_workers} workers with {num_items} items.")
+        print(
+            f"Starting {self.num_workers} workers with {num_items} items."
+            f" The progress bar is only updated when a worker finishes."
+        )
 
         if self.input_dir is None and self.src_resolver is not None and self.input_dir:
             self.input_dir = self.src_resolver(self.input_dir)


Borda (approved):

src/litdata/processing/data_processor.py
```diff
@@ -931,7 +931,10 @@ def run(self, data_recipe: DataRecipe) -> None:
 
         self._cleanup_cache()
 
-        print(f"Starting {self.num_workers} workers with {num_items} items.")
+        print(
+            f"Starting {self.num_workers} workers with {num_items} items."
+            f" The progress bar is only updated when a worker finishes."
```
tchaton, line 936: This isn't correct. It gets updated when an item is processed with a delay of up to 3 seconds to reduce queue pressure.

carmocca: Do you mean every time a worker yields?

tchaton: No, yielding in itself doesn't increment the progress bar as a single item could yield a very large number of sub items.

However, we could support some kind of progress if we wanted to.

carmocca: Then can you clarify what you mean with "an item is processed"?

tchaton: The processed items are the one provided to the optimize or map as inputs.

So if you pass 10 files, each one of them yielding 1000, the progress bar has a total of 10 elements and updates when one of the file is processed.

carmocca: Let me first tell you about my use case. I have 50 files, end processing each takes around 30 minutes. By default, litdata will start 50 workers each taking one file.

Since the progress bar only updates when the file is processed (which is also when a worker finishes), I get a blank progress bar for 30 minutes and then it goes from 0 to 50 in ~1-2 minutes as all of the workers finish.

If the user of my library does not know this and doesn't have the patience, they'll likely think that it got stuck because having a progress bar that only updates at the very end is misleading.

This is why I'm suggesting that we describe how this work precisely in this message

tchaton: Yeah, great use case.

I think we should modify the progress bar mechanism to update after every yield. I think this should provide a better experience. Wdyt ?

carmocca: It might not be worth it adding this extra communication when the user can define progress bars as they want themselves (as I do in https://github.com/Lightning-AI/litdata/issues/76).

Is there a specific message here that you would accept that clarifies this or should this be closed?

PR #95, opened 2024-04-08
Add dist env detection via env vars

### LINKED ISSUE #91: Issue with StreamingDataset when not using all GPUs on host.
## 🐛 Bug
When initializing a [StreamingDataset](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataset.py) object, `_DistributedEnv.detect()` is called ([code](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataset.py#L76)) and in the `detect()` function there a world_size check ([code](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/utilities/env.py#L53-L54)). This check fails for my use case when I am not using all GPUs on a host, such that world_size is set to 6 but `torch.cuda.device_count()` will return 8. `6 % 8 != 0`, thus raising the error.

Perhaps this check failure is the intended behavior, but I do not know enough about the `litdata` repository to understand why the code should raise an error when `world_size % device_count != 0`. I would imagine that torch would run various checks when setting up a torch distributed environment, such that this check would not be needed unless it provides a particular purpose to `StreamingDataset()` or another object. If there is any insight here on (1) if this check is needed and (2) why it is needed, that would be great!

### To Reproduce
This issue came up when experimenting with LitGPT. I don't think this section or the following sections are necessary and I will leave the blank for now. Please let me know if any of these sections would help and I can add more description.

#### Code sample
Not needed at the moment.

### Expected behavior
Not needed at the moment.

### Environment
Not needed at the moment.

gkroiz:
<details>

- [X] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)
- [X] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes #91

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/utilities/env.py
+++ b/src/litdata/utilities/env.py
@@ -57,7 +57,16 @@ class _DistributedEnv:
             global_rank = 0
             num_nodes = 1
 
-        if world_size is None or world_size == -1:
+        if os.environ["WORLD_SIZE"] is not None:
+            world_size = os.environ["WORLD_SIZE"]
+
+        if os.environ["GLOBAL_RANK"] is not None:
+            global_rank = os.environ["GLOBAL_RANK"]
+
+        if os.environ["NNODES"] is not None:
+            num_nodes = os.environ["NNODES"]
+
+        if world_size is in [None, -1, 0]:
             world_size = 1
 
         return cls(world_size=world_size, global_rank=global_rank, num_nodes=num_nodes)


gkroiz:
@tchaton here is the PR based on our conversations in #91. I realized that there are several other objects that used `_DistributedEnv`, so decided not to go down the route of passing `_DistributedEnv` as an optional argument. Instead, I went with adding environment variables that can override the detected vars. 

Two things regarding this PR's precheck. 

1) I couldn't find a Contributor guideline file, is this signifcantly different than the lightning contributer guideline file?
2) I didn't find a lot of documentation worth updating, but if you know of any documentation relevant here, I would be happy to update.

tchaton (approved):

PR #96, opened 2024-04-09
Added call to setup function of serializer class to set data format

vgurev:
<details>
  <summary><b>Fixed bug. No calls were made to the setup function of serializer class. Reading failed for serializers that required setup, for instance, serializers for no_header_numpy fails</b></summary>

</details>

## What does this PR do?

We wrote a code that writes dataset with numpy arrays. Litdata StreamingDataset fails to read the data. It seems that numpy arrays were serialized as no_header_numpy that requires call to setup function to set data format in serializer class. However, that call is not anywhere in the current code. 

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/item_loader.py
+++ b/src/litdata/streaming/item_loader.py
@@ -38,6 +38,20 @@ class BaseItemLoader(ABC):
         self._chunks = chunks
         self._serializers = serializers
 
+        # setup the serializers on restart
+        for data_format in self._config["data_format"]:
+            serializer = self._serializers[self._data_format_to_key(data_format)]
+            serializer.setup(data_format)
+
+    @functools.lru_cache(maxsize=128)
+    def _data_format_to_key(self, data_format: str) -> str:
+        if ":" in data_format:
+            serialier, serializer_sub_type = data_format.split(":")
+            if serializer_sub_type in self._serializers:
+                return serializer_sub_type
+            return serialier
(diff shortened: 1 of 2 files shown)


src/litdata/streaming/item_loader.py
```diff
@@ -126,6 +126,7 @@ def deserialize(self, raw_item_data: bytes) -> "PyTree":
         for size, data_format in zip(sizes, self._config["data_format"]):
             serializer = self._serializers[self._data_format_to_key(data_format)]
             data_bytes = raw_item_data[idx : idx + size]
+            serializer.setup(data_format)
```
tchaton, line 129: We should probably do this only once to avoid losing time. Also, would you mind adding a test ?

vgurev: Yes, single call makes sense. However, single call would require quite a refactoring of the code. I suggest this as a temporary fix. I will write a test.

tchaton: It could be done on instantiation in the base class there: https://github.com/Lightning-AI/litdata/pull/96/files/49976880c0279a075697ec3bc6b47b8d7549c51d#diff-378fc82b9b3cccccce134a1610f9dd2db00ee3d87b361ad84617687fef081790R39

tchaton (approved):
Thanks for your contribution

PR #100, opened 2024-04-11
Fix `map()` failing to create dataset when `input_dir` is None 

### LINKED ISSUE #99: Dataset not created when using `map()` on data structure without file paths inside
## 🐛 Bug

Dataset does not get created when `map` is given a list of items without paths. 

### To Reproduce

1. `pip install litdata==0.2.3 lightning_sdk==0.1.3`
2. Run this script
3. Wait for job to finish
4. Restart studio
5. `ls /teamspace/datasets `

#### Code sample

```py
import requests, os
from litdata import map
from lightning_sdk import Machine


def create_files(idx, output_dir):
    with open(os.path.join(output_dir, f"{idx}.txt") , "w") as f:
        f.write(str(idx))


def main():
    # root_dir = "./data-processed"
    root_dir = "/teamspace/datasets/data-processed"
    # os.makedirs(root_dir, exist_ok=True)

    inputs = list(range(100))

    map(
        fn=create_files,
        inputs=inputs,
        output_dir=root_dir,
        num_workers=os.cpu_count(),
        num_nodes=2,
        machine=Machine.CPU,
    )


if __name__ == "__main__":
    main()
```

In https://github.com/Lightning-AI/litdata/pull/73 a condition `and self.input_dir.path` was added:
https://github.com/Lightning-AI/litdata/blame/58f7aeb5836383ff839b4d966462c568aa6e7435/src/litdata/processing/data_processor.py#L1022
The assumption there was that map gets a datastructure of file paths, but that's not always true. For example, map could be called on a list of URLs to download and process.

awaelchli:
Fixes #99 

Not proud of the test that I added. It basically only tests that the `create_dataset` function is called. The test fails on main, but I'm looking for suggestions to make this more reliable.


src/litdata/processing/data_processor.py
```diff
@@ -1019,7 +1019,7 @@ def run(self, data_recipe: DataRecipe) -> None:
         print("Workers are finished.")
         result = data_recipe._done(len(user_items), self.delete_cached_files, self.output_dir)
 
-        if num_nodes == node_rank + 1 and self.output_dir.url and _IS_IN_STUDIO and self.input_dir.path:
+        if num_nodes == node_rank + 1 and self.output_dir.url and _IS_IN_STUDIO:  # and self.input_dir.path:
```
tchaton, line 1022: I think we should add (self.output_dir.path && self.output_dir.startswith("/teamspace/datasets"))

tchaton (approved):

src/litdata/processing/data_processor.py
awaelchli: couldn't the output dir also be an s3 url? Isn't the output_dir.url already resolved to the S3 location, even when it is a `teamspace/datasets` path?

tchaton: Yes, but I think this would be true if you provided the path to an s3 connection and create an empty dataset

PR #131, opened 2024-05-18
Fix empty tensor deserialization

enrico-stauss:
## What does this PR do?

When trying to deserialize tensors that were empty in the first place, the torch.frombuffer` method would fail due to an empty buffer. This PR fixes the issue by simply checking the buffer length and if the latter is 0, initializing a fresh tensor with no data but the correct shape and dtype.

## PR review

## Did you have fun?


tchaton (approved):
Looks great !

Issue #101, opened 2024-04-13
Question: is there a plan to support streaming from GCS?


dnnspark:

tchaton:
Hey @dnnspark,

Yes, this is quite simple to add. Simply needs to add the downloader.

dnnspark:
Thanks @tchaton, do you have an idea when it's going to land (even very rough estimate)?

tchaton:
Hey @dnnspark,

If you are willing to give it a try, I can look into it this week.

dnnspark:
Sorry for the late @tchaton

I'm willing to try! But it's not blocking at the moment, so I will stay tuned about the GCS support (it will be very helpful if you ping on this thread once it's ready).

One thing I notice is that `optimize()` function assumes the data is stored on local disk (at least in the example). In my case, the raw data is at GCS (because it's too large). Is there a way to transform the data that is stored in the cloud, and save the transformed data to the cloud without having to download the entire data?

tchaton:
> Sorry for the late @tchaton
> 
> I'm willing to try! But it's not blocking at the moment, so I will stay tuned about the GCS support (it will be very helpful if you ping on this thread once it's ready).
> 
> One thing I notice is that `optimize()` function assumes the data is stored on local disk (at least in the example). In my case, the raw data is at GCS (because it's too large). Is there a way to transform the data that is stored in the cloud, and save the transformed data to the cloud without having to download the entire data?

Yes. that's why this library was built :) But I would need to add GCS support for it ;) I will try to prioritize it.

tchaton:
@dnnspark GCS support was merged.

PR #154, opened 2024-06-03
compatibility with downloading data from gcp

### LINKED ISSUE #94: GCSFuse mount + Vertex AI custom training jobs support
## 🚀 Feature

DDP on models in GCP with data stored in GCS

### Question

Has `litdata` being tested in the case you want to train on GCP (Vertex AI) instead of just storing and streaming the data from GCS?

### Motivation

I've been trying to set up ddp on GCP (Vertex AI) using lighting Fabric with my data being stored in GCS, when you use GCS and Vertex AI you can "mount" a bucket to the instance(s)/containers running in their infrastructure using [GCSFuse](https://cloud.google.com/storage/docs/gcsfuse-integrations), in such case the only thing you need to do is replacing `gs://` by `/gcs/` and the bucket acts as a file system, does `litdata` have been tested under such setting? will it work ?

I tried already with `mosaicml-streaming` and I have ran into lots of throughput issues, that lead to data starvation in the multi-node setting

I wrote this cluster environment to use configure the cluster on vertex AI + Fabric:

```python
import os
import json

from lightning.fabric.plugins.environments.lightning import LightningEnvironment


class VertexAICluster(LightningEnvironment):
    """
    Configures distributed training on a vertex ai custom training job,
    ex:
        Consider a cluster with 3 nodes, each composed of 2 gpus

        The "cluster" key in CLUSTER_SPEC will be:
            {
                'workerpool0': ['cmle-training-workerpool0-d604929a6a-0:2222'],
                'workerpool1': [
                                'cmle-training-workerpool1-d604929a6a-0:2222',
                                'cmle-training-workerpool1-d604929a6a-1:2222'
                              ]
            }

        and each process scheduled will be under the "task" key, following the same example
        the three tasks will look like this:
            task0 ("chief" spawn process) -> node 0:
            {'type': 'workerpool0', 'index': 0}
            task 1 (on first node on workerpool1) -> node 1:
            {'type': 'workerpool1', 'index': 0}
            task 2 (on second node on workerpool1) -> node 2:
            {'type': 'workerpool1', 'index': 1}
    """

    def __init__(self):
        super().__init__()
        self.cluster_spec = json.loads(os.environ['CLUSTER_SPEC'])

    @property
    def main_address(self) -> str:
        return self.cluster_spec["cluster"]["workerpool0"][0].split(':')[0]

    @property
    def main_port(self) -> int:
        """Set common fixed MASTER_PORT port across processes
        """
        return int(self.cluster_spec["cluster"]["workerpool0"][0].split(':')[1])

    def node_rank(self) -> int:
        task = self.cluster_spec["task"]
        if task["type"] == "workerpool0":
            return 0
        else:
            return task["index"] + 1
```
do I need to set up some other `env` variables if I wanted to test `litdata`?, do the ones defined here in this cluster environment are used by `litdata`, in `mosaicml` you had to configure [these](https://docs.mosaicml.com/projects/streaming/en/stable/fundamentals/environments.html#)

dangthatsright:
Similar to the S3 data downloading, this works for GCP. I've tested it with my own data but not sure how I would set up tests for this repo.

The link to the contributor guideline is dead btw.

If there are specific places you would like me to update, please let me know.

<details>

</details>

## What does this PR do?

Fixes #94 #101

## PR review

## Did you have fun?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,3 @@ filelock
 numpy
 boto3[crt]
 requests
-google-cloud-storage
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ filelock
 numpy
 boto3[crt]
 requests
+google-cloud-storage
--- a/src/litdata/streaming/downloader.py
+++ b/src/litdata/streaming/downloader.py
@@ -19,10 +19,16 @@ from typing import Any, Dict, List
 from urllib import parse
 
 from filelock import FileLock, Timeout
-from google.cloud import storage
 
-from litdata.constants import _INDEX_FILENAME
+from litdata.constants import _GOOGLE_STORAGE_AVAILABLE, _INDEX_FILENAME
 from litdata.streaming.client import S3Client
+
+if _GOOGLE_STORAGE_AVAILABLE:
+    from google.cloud import storage
+else:
+
+    class storage:
+        Client = None
 
 
 class Downloader(ABC):
--- a/src/litdata/streaming/downloader.py
+++ b/src/litdata/streaming/downloader.py
@@ -27,7 +27,7 @@ if _GOOGLE_STORAGE_AVAILABLE:
     from google.cloud import storage
 else:
 
-    class storage:
+    class storage:  # type: ignore
         Client = None
 
 
--- a/src/litdata/streaming/downloader.py
+++ b/src/litdata/streaming/downloader.py
@@ -19,6 +19,7 @@ from typing import Any, Dict, List
 from urllib import parse
 
 from filelock import FileLock, Timeout
+from google.cloud import storage
 
 from litdata.constants import _INDEX_FILENAME
 from litdata.streaming.client import S3Client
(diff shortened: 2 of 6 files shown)


src/litdata/streaming/downloader.py
```diff
@@ -87,6 +88,36 @@ def download_file(self, remote_filepath: str, local_filepath: str) -> None:
             pass
 
 
+class GCPDownloader(Downloader):
```
tchaton, line 91: Can you add a test where you mock the storage module ?

dangthatsright: added, lmk if that's what you want. i also see something in test/streaming/test_dataset.py breaking but i can't reproduce it. lmk if it's something i introduced that broke something

requirements.txt
```diff
@@ -3,3 +3,4 @@ filelock
 numpy
 boto3[crt]
 requests
+google-cloud-storage
```
tchaton, line 6: Can we move this to extras? This shouldn't be a default

tests/streaming/test_downloader.py
```diff
@@ -13,6 +20,31 @@ def test_s3_downloader_fast(tmpdir, monkeypatch):
     popen_mock.wait.assert_called()
 
 
+def test_gcp_downloader(tmpdir, monkeypatch):
```
tchaton, line 23: Looks good.

src/litdata/streaming/downloader.py
tchaton: Looks good.

requirements.txt
dangthatsright: sounds good, i figured boto3 was here as well so i added it here

tchaton: Yes, we will make it first class citizen when there is more demand for it

tchaton:
Hey @dangthatsright. I have fixed the failure in the test. When you import an optional package, you need to make sure the logic is protected by a boolean to avoid failures.

I wondered if you would be keen on contributing the same feature for the `optimize` method, so users can generate their data and push directly to gcs.

Something like this:

```python

optimize(
    ....
    output_dir="gcs://....",
)

PR #155, opened 2024-06-05
Flexibility to set device number < total device count

### LINKED ISSUE #152: Set number of devices
This line raises an error if not all devices are used for training, for example `devices = [0]` in a machine with 4 GPUs.

The flexibility around choosing number of `devices` would be beneficial for debugging, especially since using `ddp` with multiple GPUs and `breakpoints` often gives erratic behaviour.

https://github.com/Lightning-AI/litdata/blob/cc50222aeeaf725af67fa9a443ed6456edf091f9/src/litdata/utilities/env.py#L53

Could we get this fixed? Thanks!

cc: @tchaton

yhl48:
<details>

- [X] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)

</details>

## What does this PR do?

Fixes #152. However, this implementation is not fully correct, it will break if there is more than one node.

### DIFF
--- a/src/litdata/utilities/env.py
+++ b/src/litdata/utilities/env.py
@@ -49,11 +49,15 @@ class _DistributedEnv:
             world_size = torch.distributed.get_world_size()
             global_rank = torch.distributed.get_rank()
             # Note: On multi node CPU, the number of nodes won't be correct.
-            if torch.cuda.is_available():
-                if world_size // torch.cuda.device_count() >= 1:
-                    num_nodes = world_size // torch.cuda.device_count()
-                else:
-                    num_nodes = 1
+            if torch.cuda.is_available() and world_size // torch.cuda.device_count() >= 1:
+                num_nodes = world_size // torch.cuda.device_count()
+            else:
+                num_nodes = 1
+
+            # If you are using multiple nodes, we assume you are using all the GPUs.
+            # On single node, a user can be using only a few GPUs of the node.
+            if torch.cuda.is_available() and num_nodes >= 1 and world_size % torch.cuda.device_count() != 0:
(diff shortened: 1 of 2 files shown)


yhl48:
@tchaton is there a way to get the `devices` variable?

yhl48:
the problem might be more complicated than I thought, usually `devices` is set in the lightning `Trainer` class, but `_DistributedEnv` is called before that

justusschock (approved):

src/litdata/utilities/env.py
```diff
@@ -49,8 +49,14 @@ def detect(cls) -> "_DistributedEnv":
...
-            num_nodes = world_size // torch.cuda.device_count() if torch.cuda.is_available() else world_size
-            if torch.cuda.is_available() and world_size % torch.cuda.device_count() != 0:
+            if torch.cuda.is_available() and world_size // torch.cuda.device_count() >= 1:
+                num_nodes = world_size // torch.cuda.device_count()
+            else:
+                num_nodes = 1
+
+            # If you are using multiple nodes, we assume you are using all the GPUs.
+            # On single node, a user can be using only a few GPUs of the node.
+            if torch.cuda.is_available() and num_nodes >= 1 and world_size % torch.cuda.device_count() != 0:
```
yhl48, line 59: I think this line would still fail, the issue is `world_size % torch.device_count` is not necessarily 0

tchaton: This is gated by `num_nodes >= 1`. So this is checked only for multi nodes.

tchaton: Yes, good catch ! Can you push it ?

tchaton (approved):

PR #161, opened 2024-06-10
Add feature to slice, subsample and split dataset

### LINKED ISSUE #135: Subsample StreamingDataset
## 🚀 Feature
`StreamingDataset` has enabled fast data reading, which is amazing when we have a large dataset. However, currently, it does not support reading just a fraction of data, and simple methods such as slicing the dataset does not work with `StreamingDataset`.

Could we have a feature to support this? Thanks. 

cc: @tchaton

### LINKED ISSUE #145: Support splitting datasets
## 🚀 Feature

Support splitting of datasets

### Motivation

My dataset doesn't have train/val/test split and I get `RuntimeError: The provided dataset should be either an instance of StreamingDataset or CombinedStreamingDataset. Found <torch.utils.data.dataset.Subset object>` when using random splitting. I would like this workflow supported

### Pitch

The approach above is compatible with streaming datasets

### Alternatives

Pre splitting the dataset

### Additional context

- A StreamingDataset is described by its index.json. The chunks are loaded from the file into a list there: https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/config.py#L56
[3:58](https://pytorch-lightning.slack.com/archives/D05N141TKF0/p1716994719839439)
- We could have a utility that takes a StreamingDataset and return multiple ones with sliced chunks list.
- Overlaps https://github.com/Lightning-AI/litdata/issues/135

deependujha:
<details>

- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lit-data/blob/main/.github/CONTRIBUTING.md), Pull Request section?

</details>

## What does this PR do?

Fixes #135 & fixes #145.

1. Adds support to **slice StreamingDataset**
Screenshot from 2024-06-05 13-20-06

2. Adds support to **Subsample StreamingDataset** 
Screenshot from 2024-06-08 01-29-54

3. Adds support to **train_test_split StreamingDataset** 
Screenshot from 2024-06-08 01-29-18

## PR review

## Did you have fun?


src/litdata/streaming/config.py
```diff
@@ -40,24 +42,28 @@ def __init__(
...
+        self._chunks = chunks
         self._remote_dir = remote_dir
         self._item_loader = item_loader or PyTreeLoader()
 
         with open(os.path.join(self._cache_dir, _INDEX_FILENAME)) as f:
             data = json.load(f)
             self._config = data["config"]
             self._validate_item_loader()
-            self._chunks.extend(data["chunks"])
+            if chunks is None:
```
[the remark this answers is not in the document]
deependujha: Got it. So, I'm thinking of adding an else block to check if the passed chunk is the subset of json file chunk. If not, raise an error.

src/litdata/utilities/train_test_split.py
```diff
@@ -0,0 +1,135 @@
...
+    my_new_chunk_list, my_new_roi_list = split_modify_chunk_and_roi(original_chunk, splits)
+
+    for i, curr_new_dataset in enumerate(my_datasets):
+        curr_new_dataset.chunks = my_new_chunk_list[i]
+        curr_new_dataset.region_of_interest = my_new_roi_list[i]
+
+    return my_datasets
+
+
+def is_dataset_subsample(chunks: List[Dict[str, Any]], region_of_interest: List[Tuple[int, int]]) -> bool:
```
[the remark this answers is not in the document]
deependujha: sure.

src/litdata/streaming/config.py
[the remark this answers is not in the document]
deependujha: Makes sense.

tests/utilities/test_shuffle.py
```diff
@@ -22,7 +22,16 @@ def test_intra_node_chunk_shuffle():
 
 def test_associate_chunks_and_internals_to_ranks():
     indexes = [0, 1, 2, 3, 4, 5, 6, 7]
-    chunk_intervals = [[0, 50], [0, 50], [0, 50], [0, 50], [0, 50], [0, 50], [0, 50], [0, 50]]
+    chunk_intervals = [
+        [0, 0, 50, 50],
```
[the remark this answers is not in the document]
deependujha: this is making things unnecessarily complex, imo. If I chose this syntax:
```python
chunk_intervals = [
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
    ]
 ```
 Then, I'm required to implement `__eq__` and `__getitem__` methods and it still having difficulty comparing splits (as they won't be an instance of interval).
 
 The easier way is to:
 ```
  chunk_intervals = [
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
        Interval(0, 0, 50, 50),
]
  chunk_intervals = [[chnk_interval[i] for i in range(4)] for chnk_interval in chunk_intervals]
```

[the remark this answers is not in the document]
tchaton: Sounds good, but I can imagine next contributions to be confused a list with 4 items. Let's keep it as such. I might refactor in another PR

src/litdata/streaming/cache.py
```diff
@@ -46,6 +48,8 @@ def __init__(
 
         Arguments:
             input_dir: The path to where the chunks will be or are stored.
+            chunks: The chunks that were read from `input_dir/index.json` file.
+            region_of_interest: List of tuples of {start,end} of region of interest for each chunk.
```
yhl48, line 52: nit: `(start, end)`

src/litdata/utilities/train_test_split.py
```diff
@@ -0,0 +1,135 @@
...
+    return all_samples, lst
+
+
+def split_modify_chunk_and_roi(
+    chunk_list: List[Dict[str, Any]], splits: List[float]
+) -> Tuple[List[List[Dict[str, Any]]], List[List[Tuple[int, int]]]]:
+    chunk_size = chunk_list[0]["chunk_size"]
+    total_chunk_length = len(chunk_list) * chunk_size
+    each_split_item_count = [int(total_chunk_length * split) for split in splits]
+    complete_chunk_overlap_list = [int(total_chunk_length * split) // chunk_size for split in splits]
```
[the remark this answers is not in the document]
yhl48: `int(total_chunk_length * split)` can be retrieved from `each_split_item_count` so you might wanna avoid computing that again

src/litdata/utilities/train_test_split.py
```diff
@@ -0,0 +1,135 @@
...
+        lst = [item for item in lst if item not in sample]
+
+    return all_samples, lst
+
+
+def split_modify_chunk_and_roi(
+    chunk_list: List[Dict[str, Any]], splits: List[float]
+) -> Tuple[List[List[Dict[str, Any]]], List[List[Tuple[int, int]]]]:
+    chunk_size = chunk_list[0]["chunk_size"]
+    total_chunk_length = len(chunk_list) * chunk_size
```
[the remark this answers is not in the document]
yhl48: I think `chunk_size = chunk_list[0]["chunk_size"]` can be removed since the assumption about fixed `chunk_size` is flawed

src/litdata/streaming/config.py
```diff
@@ -223,3 +241,26 @@ def _validate_item_loader(self) -> None:
...
+    # if any idx of my_subsampled_chunks is None, means,
+    # some elements in subsampled_files were not actually part of chunks
+    # raise error
+    if any(not _my_subsampled_chunk for _my_subsampled_chunk in my_subsampled_chunks):
+        raise ValueError(
+            "Mismatch in subsampled files and the chunks loaded",
+            "Make sure subsampled chunks are actually part of the original chunk",
+        )
+
+    return my_subsampled_chunks
```
tchaton, line 266: No need to put `my_` in front of all variables. Let's remove those.

src/litdata/streaming/dataset.py
```diff
@@ -62,15 +62,24 @@ def __init__(
...
             max_cache_size: The maximum cache size used by the StreamingDataset.
+            subsample: Float representing fraction of the dataset to be randomly sampled (e.g., 0.1 => 10% of dataset).
 
         """
         super().__init__()
         if not isinstance(shuffle, bool):
             raise ValueError(f"Shuffle should be a boolean. Found {shuffle}")
 
+        if not isinstance(subsample, float) or subsample < 0 or subsample > 1:
+            raise ValueError("subsample must be a float with value between 0 & 1.")
```
tchaton, line 74:
```suggestion
            raise ValueError("subsample must be a float with value between 0 and 1.")
```

src/litdata/streaming/dataset.py
```diff
@@ -62,15 +62,24 @@ def __init__(
...
             serializers: The serializers used to serialize and deserialize the chunks.
             max_cache_size: The maximum cache size used by the StreamingDataset.
+            subsample: Float representing fraction of the dataset to be randomly sampled (e.g., 0.1 => 10% of dataset).
 
         """
         super().__init__()
         if not isinstance(shuffle, bool):
             raise ValueError(f"Shuffle should be a boolean. Found {shuffle}")
 
+        if not isinstance(subsample, float) or subsample < 0 or subsample > 1:
```
tchaton, line 73:
```suggestion
        if not isinstance(subsample, float) or subsample <= 0 or subsample > 1:
```

If doesn't make sense to request none of it ;)

src/litdata/utilities/dataset_utilities.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def generate_roi(chunks: List[Dict[str, Any]], item_loader: Optional[BaseItemLoader] = None) -> List[Tuple[int, int]]:
+    "Generates default region_of_interest for chunks."
+    my_roi = []
+
+    if isinstance(item_loader, TokensLoader):
+        for idx, chunk in enumerate(chunks):
+            dim = chunk["dim"]
+            num_blocks = dim // item_loader._block_size
```
tchaton, line 104: This isn't correct. The chunk_size is `dim` as it is concatenated tensors.  I believe we are loading `item_loader._block_size * dim // item_loader._block_size` but worth to double check.

src/litdata/utilities/dataset_utilities.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # create a (chunk_start, chunk_end) list to indicate our subsample from where we can read.
+    my_roi = generate_roi(original_chunks, item_loader)
+
+    # shuffle lists together
+    if shuffle and not math.isclose(subsample, 1.0):
+        # checking if subsample is 1, as if user wants complete data, then let, shuffler and sampler do the work
+        original_chunks, my_roi = shuffle_lists_together(original_chunks, my_roi, seed)
+
+    target = int(sum([roi[1] - roi[0] for roi in my_roi]) * subsample)
```
tchaton, line 71:
```suggestion
    num_items_to_subsample = int(sum([roi[1] - roi[0] for roi in my_roi]) * subsample)
```

tchaton: better to be explicit on variable to make the code more readable.

src/litdata/streaming/item_loader.py
```diff
@@ -85,14 +97,19 @@ class PyTreeLoader(BaseItemLoader):
     def __init__(self) -> None:
         self._chunk_filepaths: Dict[str, bool] = {}
 
-    def generate_intervals(self) -> List[Tuple[int, int]]:
+    def generate_intervals(self) -> List[Tuple[int, int, int, int]]:
```
tchaton, line 100: I think we could have avoided to leak sub-sampling entirely into the item loader by getting the intervals and then applying sub-sampling to them.

src/litdata/streaming/config.py
deependujha: Sure.

src/litdata/utilities/dataset_utilities.py
```diff
@@ -0,0 +1,113 @@
...
+    if "LIGHTNING_CLUSTER_ID" not in os.environ or "LIGHTNING_CLOUD_PROJECT_ID" not in os.environ:
+        cache_dir = os.path.join(_DEFAULT_CACHE_DIR, hash_object.hexdigest())
+        os.makedirs(cache_dir, exist_ok=True)
+        return cache_dir
+    cache_dir = os.path.join("/cache", "chunks", hash_object.hexdigest())
+    os.makedirs(cache_dir, exist_ok=True)
+    return cache_dir
+
+
+def generate_roi(chunks: List[Dict[str, Any]], item_loader: Optional[BaseItemLoader] = None) -> List[Tuple[int, int]]:
```
tchaton, line 97: This is `item_loader.generate_intervals()`. No need to re-implement it. This can lead to issues in the future if they get out of sync.

src/litdata/utilities/train_test_split.py
```diff
@@ -0,0 +1,85 @@
...
+        raise ValueError("Couldn't load original chunk file.")
+
+    my_datasets = [deepcopy(streaming_dataset) for _ in splits]
+
+    dataset_length = sum([my_roi[1] - my_roi[0] for my_roi in dummy_subsampled_roi])
+
+    for i, split in enumerate(splits):
+        item_count = int(dataset_length * split)
+
+        random.seed(seed)
```
tchaton, line 74: Better to create a local random seed sampler to avoid changing the seed of the user. This can lead to reproducibility issue.

src/litdata/streaming/dataset.py
```diff
@@ -62,15 +62,24 @@ def __init__(
...
+            raise ValueError("subsample must be a float with value between 0 & 1.")
+
         input_dir = _resolve_dir(input_dir)
 
         self.input_dir = input_dir
+        self.chunks: List[Dict[str, Any]] = []
+        self.region_of_interest: List[Tuple[int, int]] = []
+        self.subsampled_files, self.region_of_interest = subsample_streaming_dataset(
```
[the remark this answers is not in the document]
deependujha: Sorry, I forgot to modify:
```diff
- self.chunks: List[Dict[str, Any]] = []
+ self.subsampled_files: List[str] = []
```

All the relevant subsampling code is in `src/litdata/utilities/dataset_utilities.py` file, and calling function `subsample_streaming_dataset` function.

I believe this already is cleaner, and makes sense. Having it in config file, is weird to me, since, what are the subsampled_files and their roi of a dataset is directly available as it's own property. 

This approach makes it possible to **split a subsampled dataset** or **subsample a splitted dataset** or whatever permutation you wish to do.

---

Now, coming to doing it only when subsample argument is provided, makes complete sense. As we are doing **unnecessary `slightly compute intensive` calculation** of getting roi and chunk_filenames (which are obviously going to be all the files).

But, if we don't do it, config file need to have two methods for loading chunks. And, also, for subsampling dataset we will need to write additional tests to test its functionality. In the current method, all passing tests relevant to dataset justify the subsampling dataset code works.

The one alternative I think is, if subsample argument is 1 (default), don't call the `subsample_streaming_dataset` and just iterate chunks and store all the files and generate `roi`. Everything else remains unchanged.

This will make code faster in the default case saving from some computation.

----

Let me know what you think about my points. Or, a better alternative.

src/litdata/utilities/dataset_utilities.py
deependujha: I copy pasted this code from `item_loader file TokenLoader class`. I didn't understand, what exactly was this.

Is this what are you saying? I don't understand.
```python
if isinstance(item_loader, TokensLoader):
        for idx, chunk in enumerate(chunks):
            dim = chunk["dim"]
            num_blocks = dim // item_loader._block_size
            assert item_loader._block_size * dim // item_loader._block_size
            end = num_blocks
            roi.append((0, end))
```

Can you please write code for it?

src/litdata/streaming/item_loader.py
deependujha: While saving the dataset's state dict, we might lose information on which are the relevant intervals. 
```python
"item_loader": self.item_loader.state_dict() if self.item_loader else None,
```
And only TokenLoader saves `block_size` info. But, we can modify this.

---

Also, let's say a user subsampled a dataset (let's say 0.3), and after some EDA, she wants to train a dummy model. So, she might want to split the subsampled dataset into (0.9, 0.1) and do her training and then validate.

To support this (`split a subsampled dataset`), we need to tweak the `roi` and `subsampled_files` only. Rest all can remain as it is, and we are done with deepcopy.

If these properties are part of StreamingDataset class, we can directly access, else, we will be required to go deeper and then get the item_loader and then it's roi and relevant filenames.

I feel like, I'm trying hard not to change my code anymore, lol. But, that's not the case. This is what I genuinely feel.🫠

src/litdata/utilities/dataset_utilities.py
deependujha: But, Both are different. Original one create intervals like: `[(0,50), (50, 100), (100,150)..]` and this one creates `[(0,50), (0, 50), (0, 50)..]`.

It is so bcoz, no matter how you shuffle, but for this chunk, you can read from (X, Y).

When item_loader actually generates the interval, here's the code:

```python
def generate_intervals(self) -> List[Tuple[int, int, int, int]]:
        intervals = []
        begin = 0
        end = 0
        for idx, curr_chunk in enumerate(self._chunks):
            end += curr_chunk["chunk_size"]
           # -------------------------------------------------------------------
            start_idx, end_idx = begin, end # it will be overlap in most cases
            if self.region_of_interest is not None:
                start_idx = begin + self.region_of_interest[idx][0] # when you are allowed to ready from a subset of chunk
                end_idx = begin + self.region_of_interest[idx][1] # when you are allowed to ready from a subset of chunk
           # -------------------------------------------------------------------
            intervals.append((begin, start_idx, end_idx, end))
            begin += curr_chunk["chunk_size"]
        return intervals
```

src/litdata/utilities/train_test_split.py
deependujha: I don't get it. Can you explain what you mean?

src/litdata/streaming/item_loader.py
deependujha: Here's a new testcase added for the same:
```python
def test_split_a_subsampled_dataset(tmpdir, compression):
    cache = Cache(str(tmpdir), chunk_size=15, compression=compression)
    for i in range(1000):
        cache[i] = i
    cache.done()
    cache.merge()

    _sub_sampled_streaming_dataset = StreamingDataset(input_dir=str(tmpdir), subsample=0.3)

    assert len(_sub_sampled_streaming_dataset) == 300 # 1000 * 0.3

    _split_fraction = [0.2, 0.3, 0.4, 0.1]

    split_datasets = train_test_split(_sub_sampled_streaming_dataset, _split_fraction)

    assert all(len(split_datasets[i])==int(300*split) for i, split in enumerate(_split_fraction))
```

tchaton: Ok, sounds good with me.

src/litdata/utilities/dataset_utilities.py
tchaton: Good point.

src/litdata/utilities/train_test_split.py
tchaton: You are setting the seed globally. Here, you want to create a local random state.

Example with numpy:

```
np.random.RandomState([self.seed, num_chunks * current_epoch, chunk_index]).permutation(array).tolist()
```

deependujha:
If subsample is 1 (default or passed), expensive tabulation (optimized in its own way) isn't called. Also, using  a local random seed sampler to avoid changing the seed of the user.

The only remaining one is: [wrong chunk size with dim](https://github.com/Lightning-AI/litdata/pull/161#discussion_r1640834400)

And, `test_s3_streaming_dataset` test passes in Lightning Studio, but fails in CI.

cc: @tchaton

src/litdata/utilities/train_test_split.py
```diff
@@ -0,0 +1,115 @@
...
+        curr_chunk_filename, curr_chunk_roi, left_chunks, left_roi = subsample_filenames_and_roi(
+            subsampled_chunks, dummy_subsampled_roi, item_count
+        )
+
+        new_datasets[i].subsampled_files = curr_chunk_filename
+        new_datasets[i].region_of_interest = curr_chunk_roi
+        subsampled_chunks = left_chunks
+        dummy_subsampled_roi = left_roi
+
+    # undo all the properties associated with original dataset
```
tchaton, line 90: Let's have a `reset` function on the StreamingDataset. This would make this more reliable if we add new arguments.

tests/utilities/test_shuffle.py
```diff
@@ -21,8 +23,37 @@ def test_intra_node_chunk_shuffle():
 
 
 def test_associate_chunks_and_internals_to_ranks():
+    @dataclass
+    class Interval:
```
tchaton, line 27: Oh, at the time, I meant to use this in the Config and StreamingDataset, not in the test ;) But all good.

tchaton (commented):
Looks awesome. I will play with it and check whether it works properly in the cloud.

src/litdata/utilities/train_test_split.py
deependujha: Great idea. Done!

src/litdata/utilities/subsample.py
```diff
@@ -0,0 +1,172 @@
...
+    Returns:
+        A tuple containing the shuffled versions of list1 and list2.
+
+    """
+    # Sanity check
+    if len(list1) != len(list2):
+        raise ValueError("Lists must be of the same size.")
+
+    if random_seed_sampler is None:
+        random_seed_sampler = np.random.RandomState([42])
```
tchaton, line 25: Let's pass down the seed as an argument.

deependujha: Is this what you mean:
```python
def shuffle_lists_together(
    list1: List[Any], list2: List[Any], random_seed_sampler: Optional[np.random.RandomState] = None, seed: Optional[int] = None
) -> Tuple[List[Any], List[Any]]:
 
    if random_seed_sampler is None and seed is None:
        raise ValueError("Both random_seed_sampler and seed are none.")

    if random_seed_sampler is None and seed is None:
        raise ValueError("Both random_seed_sampler and seed have been provided. Provide either one of them only.")

    # Sanity check
    if len(list1) != len(list2):
        raise ValueError("Lists must be of the same size.")

    if random_seed_sampler is None:
        random_seed_sampler = np.random.RandomState([seed])
    # ........ rest all fine
```

lantiga (approved):
Looks great! Just minor nits on my part, let’s try this out with redpajama

src/litdata/utilities/subsample.py
tchaton: I think seed: Optional[int] = None is enough

src/litdata/streaming/config.py
```diff
@@ -223,3 +241,26 @@ def _validate_item_loader(self) -> None:
             and not isinstance(self._item_loader, TokensLoader)
         ):
             raise ValueError("Please, use Cache(..., item_loader=TokensLoader(block_size=...))")
+
+
+def load_subsampled_chunks(subsampled_files: List[str], original_chunks: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """Loads Chunks based on the basis of subsample provided."""
+    _subsampled_chunks: List[Dict[str, Any]] = [{}] * len(subsampled_files)
```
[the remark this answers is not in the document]
deependujha: God, How didn't I pay attention to that!
```python
[{} for _ in range(len(subsampled_files))]
```
That's the new way of initializing.

src/litdata/streaming/config.py
```diff
@@ -223,3 +241,26 @@ def _validate_item_loader(self) -> None:
...
+
+    assert len(_subsampled_chunks) == len(subsampled_files)
+
+    for curr_chunk in original_chunks:
+        if curr_chunk["filename"] in subsampled_files:
```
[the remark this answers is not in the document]
deependujha: sure.
```python
# Create a defaultdict with a default value of -1 (indicates filename not present)
    filename_dict = defaultdict(lambda: -1)

    # Populate the dictionary with filenames and their indices
    for index, filename in enumerate(subsampled_files):
        filename_dict[filename] = index

    for curr_chunk in original_chunks:
        if filename_dict[curr_chunk["filename"]] != -1:
            idx = filename_dict[curr_chunk["filename"]]
            _subsampled_chunks[idx] = curr_chunk

# ............. continue from here
```

src/litdata/streaming/reader.py
```diff
@@ -294,7 +306,7 @@ def get_length(self) -> int:
 
         return len(self.config)
 
-    def get_chunk_intervals(self) -> List[Tuple[int, int]]:
```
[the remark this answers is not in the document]
deependujha: Sure, @tchaton wanted the same.

Working on it.

tchaton (approved):
Awesome contribution. Thanks you :)   

image

PR #180, opened 2024-06-22
Feat: Append data to pre-optimize dataset

### LINKED ISSUE #23: Append data to pre-optimized dataset
### Description & Motivation

It's not uncommon to have to update the data one is training on or computing embeddings on etc. We could support appending data to an optimized (chunked) dataset.
Appending alone is sufficient, removal is more specific and can be performed by rerunning through the samples with a map and creating a new dataset from it.

### Pitch

When we map or optimize specifying a target location, add ability to append to existing chunks (through a `mode="append"` argument or something along these lines).

### Alternatives

The alternative is creating a new dataset and compose with the previous one during iteration. However composition is not trivial because we need to make sure we are drawing each sample once from each and then avoid bumping into StopIteration. We would need to add a specific mode to composed dataset.

If the data added is misaligned with the chunk size and appending happens often it would create a suboptimal dataset after a while, that would need to be compacted into a single one by iterating sequentially. This could be a further alternative: a utility where you pass a list of datasets and create a single dataset by iterating through all of them.

### Additional context

_No response_

cc @borda @tchaton

Moved from https://github.com/Lightning-AI/pytorch-lightning/issues/19519, submitted by @lantiga

deependujha:
<details>

</details>

## What does this PR do?

Fixes #23 

A test to understand this feature best is:
```python
# tests/processing/test_functions.py file

def test_optimize_function_modes(tmpdir):
    output_dir = tmpdir.mkdir("output")
    output_dir = str(output_dir)

    def compress(index: int) -> Tuple[int, int]:
        return (index, index ** 2)
    
    def different_compress(index: int)->  Tuple[int, int, int]:
        return (index, index ** 2, index**3)

    # none mode
    optimize(
        fn=compress,
        inputs=list(range(1, 101)),
        output_dir = output_dir,
        chunk_bytes="64MB",
    )

    my_dataset = StreamingDataset(output_dir)
    assert len(my_dataset) == 100
    assert my_dataset[:] == [(i, i**2) for i in range(1, 101)]


    # append mode
    optimize(
        fn=compress,
        mode = "append",
        inputs=list(range(101, 201)),
        output_dir=output_dir,
        chunk_bytes="64MB",
    )

    my_dataset = StreamingDataset(output_dir)
    assert len(my_dataset) == 200
    assert my_dataset[:] == [(i, i**2) for i in range(1, 201)]


    # overwrite mode
    optimize(
        fn=compress,
        mode = "overwrite",
        inputs=list(range(201, 351)),
        output_dir=output_dir,
        chunk_bytes="64MB",
    )

    my_dataset = StreamingDataset(output_dir)
    assert len(my_dataset) == 150
    assert my_dataset[:] == [(i, i**2) for i in range(201, 351)]


    # failing case
    with pytest.raises(ValueError, match="The config of the optimized dataset is different from the original one."):
        # overwrite mode
        optimize(
            fn=different_compress,
            mode = "overwrite",
            inputs=list(range(201, 351)),
            output_dir=output_dir,
            chunk_bytes="64MB",
        )

```

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/processing/functions.py
+++ b/src/litdata/processing/functions.py
@@ -14,12 +14,12 @@
 import concurrent.futures
 import inspect
 import os
-from time import time
 from datetime import datetime
 from functools import partial
 from pathlib import Path
+from time import time
 from types import FunctionType
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union, Literal
+from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple, Union
 from urllib import parse
 
 import torch
@@ -324,7 +324,7 @@ def optimize(
 
     """
     if mode is not None and mode not in ["append", "overwrite"]:
-        raise ValueError(f"The provided mode {mode} isn't supported. Use `append`, or `overwrite`, or None.")  
+        raise ValueError(f"The provided mode {mode} isn't supported. Use `append`, or `overwrite`, or None.")
 
     if mode in ["append", "overwrite"]:
         # recursively call optimize with the same arguments but output_dir/temp_dir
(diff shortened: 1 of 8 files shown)


deependujha:
It needs to be tested if it works for S3, as it requires pro account to set up s3 connection with LightningAI studio.

But, if it works, then one improvement can be done:
> If the incompatible (`different config`) datasets are `appended` or `overwritten`, the current behaviour is to do all the stuff and then in the end, try to merge/ overwrite them.
> So, the better approach will be to check for the compatibility in the beginning itself.

---

Also, tests are passing in lightning studio. I don't know why they failed here.

PR #184, opened 2024-06-26
Feat: Append data to pre-optimize dataset

### LINKED ISSUE #23: Append data to pre-optimized dataset

deependujha:
<details>

</details>

## What does this PR do?

Fixes #23 .

Append or Overwrite to pre-optimized dataset.

<details>
  <summary>Expand here for a testcase that conveys the feature best:</summary>

```python

def test_optimize(tmpdir):
    output_dir = str(tmpdir / "output_dir") # or s3 URI

    def compress(index):
        return index, index**2

    def different_compress(index):
        return index, index**2, index**3

    optimize(
        fn=compress,
        inputs=list(range(100)),
        num_workers=1,
        output_dir=output_dir,
        chunk_bytes="64MB",
    )

    ds = StreamingDataset(output_dir)

    assert len(ds) == 100
    assert ds[:] == [(i, i**2) for i in range(100)]

    with pytest.raises(RuntimeError, match="HINT: If you want to append/overwrite to the existing dataset"):
        optimize(
            fn=compress,
            inputs=list(range(100, 200)),
            num_workers=1,
            output_dir=output_dir,
            chunk_bytes="64MB",
        )

    with pytest.raises(ValueError, match="The provided `mode` should be either `append` or `overwrite`"):
        optimize(
            fn=compress,
            inputs=list(range(100, 200)),
            num_workers=1,
            output_dir=output_dir,
            chunk_bytes="64MB",
            mode="some-random-mode",
        )

    optimize(
        fn=compress,
        inputs=list(range(100, 200)),
        num_workers=3,
        output_dir=output_dir,
        chunk_bytes="64MB",
        mode="overwrite",
    )

    ds = StreamingDataset(output_dir)

    assert len(ds) == 100
    assert ds[:] == [(i, i**2) for i in range(100, 200)]

    optimize(
        fn=compress,
        inputs=list(range(200, 300)),
        num_workers=os.cpu_count(),
        output_dir=output_dir,
        chunk_bytes="64MB",
        mode="append",
    )

    ds = StreamingDataset(output_dir)

    assert len(ds) == 200
    assert ds[:] == [(i, i**2) for i in range(100, 300)]

    optimize(
        fn=compress,
        inputs=list(range(300, 400)),
        num_workers=2,
        output_dir=output_dir,
        chunk_bytes="64MB",
        mode="append",
    )

    ds = StreamingDataset(output_dir)

    assert len(ds) == 300
    assert ds[:] == [(i, i**2) for i in range(100, 400)]

    with pytest.raises(Exception, match="The config isn't consistent between chunks"):
        optimize(
            fn=different_compress,
            inputs=list(range(100, 200)),
            num_workers=1,
            output_dir=output_dir,
            chunk_bytes="64MB",
            mode="append",
        )

    ds = StreamingDataset(output_dir)

    assert len(ds) == 300
    assert ds[:] == [(i, i**2) for i in range(100, 400)]

    optimize(
        fn=different_compress,
        inputs=list(range(800, 900)),
        num_workers=1,
        output_dir=output_dir,
        chunk_bytes="64MB",
        mode="overwrite",
    )

    ds = StreamingDataset(output_dir)

    assert len(ds) == 100
    assert ds[:] == [(i, i**2, i**3) for i in range(800, 900)]

```
</details>

Above testcase is also present at `litdata/tests/processing/test_functions.py:48`

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/processing/utilities.py
+++ b/src/litdata/processing/utilities.py
@@ -236,7 +236,7 @@ def delete_index_file(output_dir: Dir) -> None:
     """Delete the index file."""
     if not isinstance(output_dir, Dir):
         raise ValueError("The provided output_dir should be a Dir object.")
-    
+
     if output_dir.url is None:
         if output_dir.path is None:
             return
--- a/src/litdata/processing/utilities.py
+++ b/src/litdata/processing/utilities.py
@@ -13,12 +13,17 @@
 
 import io
 import os
+import json
 import urllib
+import tempfile
+from urllib import parse
 from contextlib import contextmanager
 from subprocess import DEVNULL, Popen
-from typing import Any, Callable, List, Optional, Tuple, Union
+from typing import Any, Callable, List, Optional, Tuple, Union, Dict
 
 from litdata.constants import _IS_IN_STUDIO, _LIGHTNING_CLOUD_AVAILABLE
+from litdata.streaming.cache import Dir
+from litdata.constants import _INDEX_FILENAME
 
 if _LIGHTNING_CLOUD_AVAILABLE:
     from lightning_cloud.openapi import (
@@ -26,6 +31,15 @@ if _LIGHTNING_CLOUD_AVAILABLE:
     )
     from lightning_cloud.openapi.rest import ApiException
     from lightning_cloud.rest_client import LightningClient
+
+try:
+    import boto3
+    import botocore
+
+    _BOTO3_AVAILABLE = True
+except Exception:
+    _BOTO3_AVAILABLE = False
+
 
 
 def _create_dataset(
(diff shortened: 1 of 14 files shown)


src/litdata/processing/utilities.py
```diff
@@ -177,3 +189,63 @@ def _get_work_dir() -> str:
...
+        if not os.path.exists(index_file_path):
+            return None
+        with open(index_file_path) as f:
+            return json.load(f)
+
+    else:
+        # download the index file from s3, and read it
+        obj = parse.urlparse(output_dir.url)
+
+        if obj.scheme != "s3":
```
tchaton, line 212: Note for another PR. We support GCP in a minimal way, so it would be good to make this more general. As suggested in one of the issues, we could use fsspec.

tchaton (approved):

PR #210, opened 2024-07-06
Feat: adds support for reading mosaic mds written dataset

### LINKED ISSUE #195: Add support for Mosaic Streaming WDS data format
### Motivation

The format are similar and it would be nice to be able to use a Mosaic streaming dataset with LitData.

bhimrazy:
## What does this PR do?

Fixes #195.

### DIFF
--- a/src/litdata/utilities/dataset_utilities.py
+++ b/src/litdata/utilities/dataset_utilities.py
@@ -138,11 +138,11 @@ def load_index_file(input_dir: str) -> Dict[str, Any]:
         with open(index_filepath) as f:
             data = json.load(f)
 
-        if "chunks" in data:
-            return data
-        if "shards" in data:
+        if "chunks" not in data and "shards" in data:
+            # load mds shard-based index file and adapt to chunks format
             return adapt_mds_shards_to_chunks(data)
-        raise ValueError(f"Invalid index file format at {index_filepath}.")
+
+        return data
     except FileNotFoundError:
         raise FileNotFoundError(f"Index file not found at {index_filepath}.")
 
--- a/src/litdata/utilities/dataset_utilities.py
+++ b/src/litdata/utilities/dataset_utilities.py
@@ -51,9 +51,8 @@ def subsample_streaming_dataset(
 
     if os.path.exists(os.path.join(input_dir.path, _INDEX_FILENAME)):
         # load chunks from `index.json` file
-        with open(os.path.join(input_dir.path, _INDEX_FILENAME)) as f:
-            data = json.load(f)
-            original_chunks = data["chunks"]
+        data = load_index_file(input_dir.path)
+        original_chunks = data["chunks"]
     else:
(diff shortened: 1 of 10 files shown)


src/litdata/utilities/dataset_utilities.py
```diff
@@ -115,3 +114,57 @@ def generate_roi(chunks: List[Dict[str, Any]], item_loader: Optional[BaseItemLoa
...
+
+    index_filepath = os.path.join(input_dir, _INDEX_FILENAME)
+    try:
+        # load index.json file
+        with open(index_filepath) as f:
+            data = json.load(f)
+            if "chunks" not in data:
+                raise KeyError(f"'chunks' not found in the index file at {index_filepath}.")
+            return data
+    except KeyError as e:
```
deependujha, line 130: Instead of relying on try-except, the better approach will be to `use conditionals (if, elif, else)`.

If tomorrow we have to add support for another StreamingDataset, we will be required to make a nested try-except code which will be difficult to read.

Also, extract out the logic of reading mds index_json files into another function. Let's keep this function clean.

```python
# something like

if "chunks" in data:
  # litdata optimize
  return data
elif "shards" in data:
  # mds writer
  return read_mds_index_json()
else:
  ...
```

tests/streaming/test_dataset.py
```diff
@@ -995,3 +995,27 @@ def test_subsample_streaming_dataset_with_token_loader(tmpdir, monkeypatch):
...
+    from streaming import MDSWriter
+    # example taken from: https://github.com/mosaicml/streaming
+
+    # A dictionary mapping input fields to their data types
+    columns = {"image": "jpeg", "class": "int"}
+    # Shard compression, if any
+    compression = "zstd"
+    # Save the samples as shards using MDSWriter
+    with MDSWriter(out=str(tmpdir), columns=columns, compression=compression) as out:
+        for i in range(100):
```
deependujha, line 1011: Even 10 should be enough.

And, try fixing the failing test. If it continues: `@pytest.mark.skipif(sys.platform == "win32" or sys.platform == "darwin", reason="...")`. But, be sure before using this.

deependujha (commented):
Great work. Few suggestions from my side.

And update the wrong spelling of `Parallelize` in readme, which is causing pre-commit-ci to fail.

```
README.md:55: Paralellize ==> Parallelize
README.md:653: Paralellize ==> Parallelize
```

deependujha (approved):
Looks good to me. 

Added tests for checking if the mds dataset supports **`Litdata streamingDataset features`** like `train_test_split`, `subsample`, `dataloader (last_drop=True | False)`.

One test fails on mac (timeout), even though I've not changed anything in any function.

bhimrazy:
Thank you @deependujha for adding these additional tests.

tchaton (approved):
Great work ! It would be good to benchmark this once more.

Issue #260, opened 2024-07-23
If data samples contain a Python list with a variable number of elements, type inference will fail


senarvi:
## 🐛 Bug

When optimizing a dataset, [`BinaryWriter.serialize()`](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/writer.py#L167) will first flatten the sample dictionary and infer the type of each element. Then it will assume that all data samples have the same types and the same number of elements. If a list contains a variable number of elements, the type information will be incorrect for subsequent data samples. This is not documented and can cause some confusion.

### To Reproduce

When storing the data in NumPy arrays, there's no problem, because each array is considered as one element in the flattened type list. So this works:

```python
import numpy as np
from PIL import Image
import litdata as ld

def random_images(index):
    fake_images = Image.fromarray(np.random.randint(0, 256, (32, 32, 3), dtype=np.uint8))
    num_labels = np.random.randint(10)
    fake_labels = np.random.randint(10, size=num_labels)

    # use any key:value pairs
    data = {"index": index, "image": fake_images, "class": fake_labels}

    return data

if __name__ == "__main__":
    # the optimize function outputs data in an optimized format (chunked, binerized, etc...)
    ld.optimize(
        fn=random_images,                   # the function applied to each input
        inputs=list(range(1000)),           # the inputs to the function (here it's a list of numbers)
        output_dir="my_optimized_dataset",  # optimized data is stored here
        num_workers=4,                      # The number of workers on the same machine
        chunk_bytes="64MB"                  # size of each chunk
    )
```

When storing the data in a Python list, the type of each list element is inferred separately. If a list contains a variable number of elements, the type information of one sample is not useful for other samples. Now, if we add an element that has a different type (say, a string) after a variable-length list, the function will crash:

```python
def random_images(index):
    fake_images = Image.fromarray(np.random.randint(0, 256, (32, 32, 3), dtype=np.uint8))
    num_labels = np.random.randint(10)
    fake_labels = np.random.randint(10, size=num_labels).tolist()
    name = "name"

    # use any key:value pairs
    data = {"index": index, "image": fake_images, "class": fake_labels, "name": name}

    return data
```

The error:

```
Rank 0 inferred the following `['int', 'pil', 'int', 'int', 'int', 'str']` data format.                                                                                                                                                                         | 0/1000 [00:00<?, ?it/s]
Rank 1 inferred the following `['int', 'pil', 'int', 'int', 'int', 'str']` data format.
[...]

PR #264, opened 2024-07-24
Update map() and optimize() documentation

senarvi:
Fixes documentation related to #260 .

Adds notes that the types and list sizes must not changed between samples, when calling `optimize()`.

When looking at the code, I noticed that the documentation of the inputs seems to be outdated. AFAICS, inputs can be any sequence or a streaming dataloder, so I updated the type annotation and removed this from the docstring: "Each input should contain at least a valid filepath". Let me know if I missed something, or feel free to make changes in the PR.

My code editor also removed unnecessary whitespace. I guess that's good.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -10,3 +10,4 @@ pandas
 lightning
 lightning-cloud == 0.5.70 # Must be pinned to ensure compatibility
 zstd
+numpy < 2.0
--- a/README.md
+++ b/README.md
@@ -4,16 +4,16 @@
 &nbsp;
 &nbsp;
 
-**Transform datasets at scale.   
+**Transform datasets at scale.
 Optimize data for fast AI model training.**
 
 
 <pre>
 Transform                              Optimize
-  
-✅ Parallelize data processing       ✅ Stream large cloud datasets          
-✅ Create vector embeddings          ✅ Accelerate training by 20x           
-✅ Run distributed inference         ✅ Pause and resume data streaming      
+
+✅ Parallelize data processing       ✅ Stream large cloud datasets
+✅ Create vector embeddings          ✅ Accelerate training by 20x
+✅ Run distributed inference         ✅ Pause and resume data streaming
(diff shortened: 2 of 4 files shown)


deependujha:
Hi @senarvi, thanks for opting to contribute to litdata.

Btw, Can you please convert your PR into draft until it's ready for review?

senarvi:
@deependujha how do I convert a PR into a draft?

senarvi:
Does anyone have an idea why the Windows and MacOS tests fail with "Numpy is not available"?

deependujha:
> @deependujha how do I convert a PR into a draft?

Screenshot from 2024-07-24 14-41-31

Just click on convert to Draft.

---

I'll try attempting the failing tests soon.

senarvi:
Thanks @deependujha , I haven't noticed that link! That will be useful in the future. From my behalf this PR is ready. I think anything in this PR is not responsible for breaking the tests, since this doesn't include functional changes. So better to leave this open, or what do you think?

deependujha:
sure. I'm looking into it. Thanks for contributing.

tchaton (approved):

requirements/test.txt
```diff
@@ -10,3 +10,4 @@ pandas
 lightning
 lightning-cloud == 0.5.70 # Must be pinned to ensure compatibility
 zstd
+numpy < 2.0
```
deependujha, line 13: Why this though?
I believe a new PR was merged recently which added support for numpy>=2!

deependujha, line 13: Why this though?
I believe a new PR was merged recently which added support for numpy>=2!

deependujha (commented):
All failing tests were bcoz of numpy. But, recently support for numpy>=2 was added.

requirements/test.txt
tchaton: Yes.

deependujha: nice answer. 😀
for the first part or second part (hack vs correct implementation)

PR #296, opened 2024-08-01
Enforce passing item_loader when customizing underlying storage format

### LINKED ISSUE #294: StreamingDataset cannot load one-element, one-dimensional tensors
## 🐛 Bug

I'm applying Litdata's `optimize` method to a dataset of one-dimensional tensors. I get an error when trying to use a `StreamingDataset` on the optimized data. 

### To Reproduce

The following script will reproduce the error

```
from lightning.data import optimize, StreamingDataset
import torch

def process(x):
    item = torch.rand(size=(10,))
    yield (item,)

if __name__ == "__main__":
    data_dir = "litdata_bug_repro"
    optimize(
        fn=process,
        inputs=list(range(100)),
        num_workers=1,
        output_dir=data_dir,
        chunk_bytes="64MB"
    )

    dataset = StreamingDataset(data_dir)
    dataset[0]
```

Here's the error

```
litdata_bug_repro.py 20 <module>
dataset[0]

dataset.py 326 __getitem__
self.cache = self._create_cache(worker_env=self.worker_env)

dataset.py 171 _create_cache
cache._reader._try_load_config()

reader.py 223 _try_load_config
self._config = ChunksConfig.load(

config.py 251 load
return ChunksConfig(

config.py 63 __init__
self._validate_item_loader()

config.py 265 _validate_item_loader
raise ValueError("Please, use Cache(..., item_loader=TokensLoader(block_size=...))")

ValueError:
Please, use Cache(..., item_loader=TokensLoader(block_size=...))

```

### Environment

- Litdata version 0.2.21 (though I first noticed it on 0.2.16)
- Python version 3.11
- PyTorch Version 2.0.1+cpu.cxx11.abi
- OS Ubuntu

### Additional context

I've only seen the error for single-element tuples containing one-dimensional tensors. All of the following work as expected
```
def process(x):
    item = np.random.normal(size=(10,))  # one-dimensional numpy array
    yield (item,)
```

```
def process(x):
    item = torch.rand(size=(10,))
    yield (item, item)  # Multiple elements
```
```
def process(x):
    item = torch.rand(size=(10,10))  # Higher-dimensional tensor
    yield (item,)
```

tchaton:
<details>

</details>

## What does this PR do?

Until now, we had a hack where 1D tensor would be handled differentely and stored as contiguous array. I have seen several users complaining about this magic and unexpected behaviour.

WARNING: This PR is a breaking change for LLM using the TokensLoader. Now, we would need to pass the item_loader to the optimize or Cache directly to inform the underlying storage needs to be handled differently.  

If no item_loader is passed during the optimization, this default to the Pytree handler.

For LLM tokens, here is the breaking API change.

Before

```python
from litdata.streaming.item_loader import TokensLoader

optimize(...)

dataset = StreamingDataset(item_loader=TokensLoader(...))
```

Now

```python
from litdata.streaming.item_loader import TokensLoader

optimize(..., item_loader=TokensLoader())

dataset = StreamingDataset(item_loader=TokensLoader(...))
```

Fixes #294

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/processing/functions.py
+++ b/src/litdata/processing/functions.py
@@ -38,6 +38,7 @@ from litdata.processing.utilities import (
 )
 from litdata.streaming.client import S3Client
 from litdata.streaming.dataloader import StreamingDataLoader
+from litdata.streaming.item_loader import BaseItemLoader
 from litdata.streaming.resolver import (
     Dir,
     _assert_dir_has_index_file,
@@ -311,6 +312,7 @@ def optimize(
     batch_size: Optional[int] = None,
     mode: Optional[Literal["append", "overwrite"]] = None,
     use_checkpoint: bool = False,
+    item_loader: Optional[BaseItemLoader] = None,
 ) -> None:
     """This function converts a dataset into chunks, possibly in a distributed way.
 
@@ -340,6 +342,7 @@ def optimize(
             Defaults to None.
         use_checkpoint: Whether to create checkpoints while processing the data, which can be used to resume the
             processing from the last checkpoint if the process is interrupted. (`Default: False`)
+        item_loader: The item loader used by the StreamingDataset if any.
 
     """
     if mode is not None and mode not in ["append", "overwrite"]:
@@ -430,6 +433,7 @@ def optimize(
             reader=reader,
             state_dict=state_dict,
             use_checkpoint=use_checkpoint,
+            item_loader=item_loader,
         )
 
         with optimize_dns_context(True):
--- a/src/litdata/processing/functions.py
+++ b/src/litdata/processing/functions.py
@@ -343,7 +343,8 @@ def optimize(
             Defaults to None.
         use_checkpoint: Whether to create checkpoints while processing the data, which can be used to resume the
             processing from the last checkpoint if the process is interrupted. (`Default: False`)
-        item_loader: The item loader used by the StreamingDataset if any.
+        item_loader: The item loader that will be used during loading in StreamingDataset. Determines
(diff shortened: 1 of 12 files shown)


lantiga (approved):
Much cleaner and more robust. IMO the change is not confusing, one just has to use item loader on both sides.
What happens when I omit it?

tchaton:
> What happens when I (erroneously) omit it?

It would break and tell you the wrong item loader has been provided.

src/litdata/processing/functions.py
```diff
@@ -340,6 +342,7 @@ def optimize(
             Defaults to None.
         use_checkpoint: Whether to create checkpoints while processing the data, which can be used to resume the
             processing from the last checkpoint if the process is interrupted. (`Default: False`)
+        item_loader: The item loader used by the StreamingDataset if any.
```
awaelchli, line 345:
```suggestion
        item_loader: The item loader that will be used during loading in StreamingDataset. Determines
        	the format in which the data is stored and optimized for loading.
```
My proposal to change the description based on my understanding. 
To make it clear this loader is not used for loading during any of the optimization, but determines how we expect the dataset to be consumed.

awaelchli (approved):

AugustDev:
`TokensLoader` accepts `block_size`. What if each of my sample is a dictionary where some of the fields are token sequences. Should `block_size` be length of the longest sequence?

tchaton:
Hey @AugustDev. TokensLoader doesn't work on dictionary. It needs a 1D tensor.

PR #311, opened 2024-08-06
Fix StreamingDataset.get_len(num_workers=0)

### LINKED ISSUE #310: StreamingDataLoader sets num_workers to zero, which causes the dataset length to be reported incorrectly
## 🐛 Bug

I'm using LitData with PyTorch Lightning. I create a StreamingDataset and a StreamingDataLoader. I'm using 0 workers, so that I can debug errors with data loading. Lightning complains that the total data length across ranks is zero. I think the problem is that when calling `StreamingDataLoader.__len__()`, the number of workers is set to 0.

### To Reproduce

Steps to reproduce the behavior (see the scripts below):

1. Generate data.
1. Observe the dataset length before and after getting the data loader length.

#### Code sample

Generate the data:
```python
import numpy as np
from PIL import Image
import litdata as ld

def random_images(index):
    fake_images = Image.fromarray(np.random.randint(0, 256, (32, 32, 3), dtype=np.uint8))
    fake_labels = np.random.randint(10)
    return {"index": index, "image": fake_images, "class": fake_labels}

if __name__ == "__main__":
    ld.optimize(
        fn=random_images,
        inputs=list(range(1000)),
        output_dir="my_optimized_dataset",
        num_workers=4,
        chunk_bytes="64MB"
    )
```

Observe the data length:
```python
import litdata as ld

dataset = ld.StreamingDataset("my_optimized_dataset", shuffle=True)
dataloader = ld.StreamingDataLoader(dataset)
print(len(dataset))  # 1000
print(len(dataloader))  # 0
print(len(dataset))  # 0
```

### Expected behavior

The first time I call `len(dataset)`, I get the correct length, 1000.

I would expect to get the same number, when I call `len(dataloader)`, but instead I get 0. After that call, `len(dataset)` also returns 0.

### Environment

- PyTorch Version (e.g., 1.0): 2.3.1
- OS (e.g., Linux): Linux
- How you installed PyTorch (`conda`, `pip`, source): pip
- Build command you used (if compiling from source):
- Python version: 3.12.3
- CUDA/cuDNN version: 12.4
- GPU models and configuration: NVIDIA GeForce GTX 1650
- Any other relevant information:

### Additional context

[`StreamingDataset`](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataset.py#L129) considers "0 workers" (i.e. data loading in the main process), as one worker.

[`StreamingDataset.set_num_workers()`](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataset.py#L197) sets the `num_workers` attribute to 1, if the value 0 is passed.

[`StreamingDataset.get_len()`](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataset.py#L200) will not call `set_num_workers()`, but sets the attribute to 0, if the value 0 is passed.

[The default value](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataloader.py#L553) for `num_workers` in  `StreamingDataLoader` is 0, and the value 0 is saved in the `num_workers` attribute.

[`StreamingDataLoader.get_len()`](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/dataloader.py#L647) calls `StreamingDataset.get_len()`, passing the value 0.

[`Shuffle.get_len()`](https://github.com/Lightning-AI/litdata/blob/main/src/litdata/streaming/shuffle.py#L41) returns 0 when `num_workers` is 0.

senarvi:
This is how I think #310 could be fixed. This hasn't been discussed yet.

## What does this PR do?

Fixes #310 .

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/dataset.py
+++ b/src/litdata/streaming/dataset.py
@@ -198,7 +198,7 @@ class StreamingDataset(IterableDataset):
 
     def get_len(self, num_workers: int, batch_size: int) -> int:
         self.set_num_workers(num_workers)
-        self.batch_size = batch_size
+        self.set_batch_size(batch_size)
         worker_env = _WorkerEnv.detect()
         if self.shuffler is None:
             cache = self._create_cache(worker_env=worker_env)
--- a/src/litdata/streaming/dataset.py
+++ b/src/litdata/streaming/dataset.py
@@ -203,7 +203,7 @@ class StreamingDataset(IterableDataset):
         if self.shuffler is None:
             cache = self._create_cache(worker_env=worker_env)
             self.shuffler = self._create_shuffler(cache)
-        return self.shuffler.get_len(self.distributed_env, self.num_workers, batch_size, self.current_epoch)
+        return self.shuffler.get_len(self.distributed_env, self.num_workers, self.batch_size, self.current_epoch)
 
     def __iter__(self) -> "StreamingDataset":
         # When the StreamingDataset is used within map or optimize, let's refetch the distributed env.
--- a/src/litdata/streaming/dataset.py
+++ b/src/litdata/streaming/dataset.py
@@ -197,13 +197,13 @@ class StreamingDataset(IterableDataset):
         self.num_workers = num_workers or 1
 
     def get_len(self, num_workers: int, batch_size: int) -> int:
-        self.num_workers = num_workers
+        self.set_num_workers(num_workers)
         self.batch_size = batch_size
         worker_env = _WorkerEnv.detect()
         if self.shuffler is None:
             cache = self._create_cache(worker_env=worker_env)
             self.shuffler = self._create_shuffler(cache)
-        return self.shuffler.get_len(self.distributed_env, num_workers, batch_size, self.current_epoch)
+        return self.shuffler.get_len(self.distributed_env, self.num_workers, batch_size, self.current_epoch)
 
     def __iter__(self) -> "StreamingDataset":
         # When the StreamingDataset is used within map or optimize, let's refetch the distributed env.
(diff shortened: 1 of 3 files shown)


src/litdata/streaming/dataset.py
```diff
@@ -197,13 +197,13 @@ def set_num_workers(self, num_workers: int) -> None:
         self.num_workers = num_workers or 1
 
     def get_len(self, num_workers: int, batch_size: int) -> int:
-        self.num_workers = num_workers
+        self.set_num_workers(num_workers)
         self.batch_size = batch_size
```
tchaton, line 201:
```suggestion
        self.set_batch_size(batch_size)
```

src/litdata/streaming/dataset.py
```diff
@@ -197,13 +197,13 @@ def set_num_workers(self, num_workers: int) -> None:
...
         worker_env = _WorkerEnv.detect()
         if self.shuffler is None:
             cache = self._create_cache(worker_env=worker_env)
             self.shuffler = self._create_shuffler(cache)
-        return self.shuffler.get_len(self.distributed_env, num_workers, batch_size, self.current_epoch)
+        return self.shuffler.get_len(self.distributed_env, self.num_workers, batch_size, self.current_epoch)
```
tchaton, line 206:
```suggestion
        return self.shuffler.get_len(self.distributed_env, self.num_workers, self.batch_size, self.current_epoch)
```

tchaton (approved):
Thanks. Just for consistency.

tchaton:
Hey @senarvi. Would you mind adding a test ?

tchaton:
Hey @senarvi. Thanks a lot for your contribution :)

senarvi:
> Hey @senarvi. Would you mind adding a test ?

@tchaton I think adding a test is a good idea. You merged this pull request already, so I created a new one: https://github.com/Lightning-AI/litdata/pull/314

Issue #341, opened 2024-08-26
Bug: Loading compressed data fails silently (no error message, the application simply hangs up)


AugustDev:
## 🐛 Bug

I am trying to install and use litdata in clean environment, but there are few problems
1. Loading compressed data fails silently (no error message, the application simply hangs up)
2. Importing litdata gives a warning about numpy 1.x
3. Loading the data gives `RuntimeError: Could not infer dtype of numpy.int64`

```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.2 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "/Users/wpc/Projects/cfdx/data/main.py", line 1, in <module>
    from litdata import StreamingDataset, StreamingDataLoader, optimize
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/litdata/__init__.py", line 16, in <module>
    from litdata.processing.functions import map, merge_datasets, optimize, walk
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/litdata/processing/functions.py", line 28, in <module>
    import torch
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/torch/__init__.py", line 1477, in <module>
    from .functional import *  # noqa: F403
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/torch/functional.py", line 9, in <module>
    import torch.nn.functional as F
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/torch/nn/__init__.py", line 1, in <module>
    from .modules import *  # noqa: F403
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/torch/nn/modules/__init__.py", line 35, in <module>
    from .transformer import TransformerEncoder, TransformerDecoder, \
  File "/Users/wpc/Projects/cfdx/data/venv3/lib/python3.9/site-packages/torch/nn/modules/transformer.py", line 20, in <module>
[...]

PR #342, opened 2024-08-27
Fix: Ensure Compression Algorithm is Installed Before Reading Compressed Data

bhimrazy:
<details>

- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lit-data/blob/main/.github/CONTRIBUTING.md), Pull Request section?

</details>

## What does this PR do?
- This PR addresses the first bug from issue #341 by raising a ValueError if the data being read is compressed but the compression algorithm is not installed.

> Added the same version of the check as seen in [L93 from `writer.py`](https://github.com/Lightning-AI/litdata/blob/ee4ff912b07eb8e8d9be80d6d20f134a46c0005a/src/litdata/streaming/writer.py#L93).

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/config.py
+++ b/src/litdata/streaming/config.py
@@ -85,11 +85,12 @@ class ChunksConfig:
 
         if self._compressor_name:
             if len(_COMPRESSORS) == 0:
-                raise ValueError("No compresion algorithms are installed.")
-
+                raise ValueError(
+                    "No compression algorithms are installed. To use zstd compression,  run `pip install zstd`."
+                )
             if self._compressor_name not in _COMPRESSORS:
                 raise ValueError(
-                    f"The provided compression {self._compressor_name} isn't available in {sorted(_COMPRESSORS)}"
+                    f"The provided compression {self._compressor_name} isn't available in {sorted(_COMPRESSORS)}",
                 )
             self._compressor = _COMPRESSORS[self._compressor_name]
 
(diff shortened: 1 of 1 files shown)


src/litdata/streaming/config.py
```diff
@@ -83,7 +83,14 @@ def __init__(
         self._compressor_name = self._config["compression"]
         self._compressor: Optional[Compressor] = None
 
-        if self._compressor_name in _COMPRESSORS:
+        if self._compressor_name:
+            if len(_COMPRESSORS) == 0:
+                raise ValueError("No compresion algorithms are installed.")
+
+            if self._compressor_name not in _COMPRESSORS:
+                raise ValueError(
```
deependujha, line 91: Btw, can we try to install the required compression algorithm rather than raising ValueError?

Something like, 
```python
print("No compression algorithms are installed")
print("Installing ....")
```

what do you think?

bhimrazy: Thank you for the review and suggestion @deependujha. That could indeed be a helpful approach. 

However, I’m wondering if it is better to leave the installation process to the user, possibly providing them with clear instructions or recommendations on how to proceed with the installation.

AugustDev:
Looks good!

AugustDev:
Really nice!

deependujha (approved):

PR #348, opened 2024-09-01
Feat: Using fsspec to download files

### LINKED ISSUE #181: Using fsspec to download files
## 🚀 Feature

I was wondering if you planned on using [fsspec](https://github.com/fsspec/filesystem_spec) for downloading file. Since fsspec is already compatible with many cloud provider it would make lit-data streaming compatible with even more cloud.

### Motivation

Right now there is one downloader class for each provider, fsspec would allow to abstract it away, reducing the amount of cloud provider specific code and allowing supporting more out of the box

PS: love the work on lit data

deependujha:
<details>

</details>

## What does this PR do?

Fixes #181 

- Basic setup and working for fsspec done and tested for S3 & GS.

- Failing tests will be modified to match fsspec tests after approval.

- Requirements.txt file will also be optimized after approval.

---

## A small script to test

```python
import litdata as ld

# modify it to azure or gcp endpoint, and give it a try (tested for s3 & gs)
s3_uri = 's3://my-dummy-bucket-litdata/deep_data/' 

dataset = ld.StreamingDataset(s3_uri, shuffle=True)
dataloader = ld.StreamingDataLoader(dataset)

for sample in dataloader:
    print(sample)
    print("="*80)
print("\nAll done\n")
```

## PR review

## Did you have fun?


src/litdata/streaming/downloader.py
```diff
@@ -188,10 +198,42 @@ def download_file(self, remote_filepath: str, local_filepath: str) -> None:
...
+_DOWNLOADERS = {
+    "s3://": "s3",
+    "gs://": "gs",
+    "azure://": "abfs",
+    "local:": "file",
+    "": "file",
+}
+
+
+class FsspecDownloader(Downloader):
```
deependujha, line 210: this class can replace all others. And, we can support many more providers, like, huggingface, dropbox, DVC, and many others.

deependujha: One con that I can see is:

- We might lose on the faster S3 bcoz of not utilizing `s5cmd` (if fsspec doesn't use it internally).

Or we can handle this special case separately.

src/litdata/streaming/downloader.py
```diff
@@ -26,8 +27,17 @@
...
+        cloud_provider: str,
+        remote_dir: str,
+        cache_dir: str,
+        chunks: List[Dict[str, Any]],
+        storage_options: Optional[Dict] = {},
     ):
+        print("-" * 100)
+        print(f"{cloud_provider=}")
+        print("-" * 100)
+        self.fs = fsspec.filesystem(cloud_provider)
```
tchaton, line 40: How do you provide the storage options to fsspec ?

src/litdata/streaming/downloader.py
```diff
@@ -188,10 +198,42 @@ def download_file(self, remote_filepath: str, local_filepath: str) -> None:
 }
 
 
+_DOWNLOADERS = {
```
tchaton, line 201: We probably don't need this mapping.

src/litdata/streaming/downloader.py
tchaton: Yes, we can probably use it only if installed.

src/litdata/streaming/downloader.py
deependujha: To be honest, I don't know much about storage options.

I'll look into s3 & gs docs to see different storage options available. And then will look back into fsspec docs to see how to pass them.

src/litdata/streaming/downloader.py
tchaton: We can probably keep only the ones that requires mapping then.

src/litdata/streaming/downloader.py
deependujha: cool, so am I on the right track?

If yes, then I first need to figure out the `storage option` stuff and test on all three providers. And, then ask for another round of review.

src/litdata/streaming/downloader.py
deependujha: https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.filesystem

Screenshot from 2024-09-02 19-07-45

Probably this one.

deependujha: It works. But, there's slight mismatch in storage options and what it expects. Independent of our code, but user need to be aware of the `storage_options`.

```python
storage_options = {
    "key": "MY_AWS_ACCESS_KEY",
    # "aws_access_key_id": "MY_AWS_ACCESS_KEY",

   "secret": "MY_AWS_SECRET_KEY",
   #  "aws_secret_access_key": "MY_AWS_SECRET_KEY",
}
```

expects `key` instead of `aws_access_key_id` and `secret` instead of `aws_secret_access_key`.

<details>
  <summary>Complete code</summary>
  
  ```python
import litdata as ld

s3_uri = 's3://my-dummy-bucket-litdata/my_litdata/'

storage_options = {
    "key": "MY_AWS_ACCESS_KEY",
    # "aws_access_key_id": "MY_AWS_ACCESS_KEY",

   "secret": "MY_AWS_SECRET_KEY",
   #  "aws_secret_access_key": "MY_AWS_SECRET_KEY",
}

dataset = ld.StreamingDataset(s3_uri, shuffle=True, storage_options=storage_options)
dataloader = ld.StreamingDataLoader(dataset)

for sample in dataloader:
    print(sample)
    print("="*80)
print("\nAll done\n")

  ```
</details>

---

IMO, this happens bcoz of the way the [S3FS class, init function & set_session works](https://github.com/fsspec/s3fs/blob/2763a083e55d64abded49269daf098be86b4a961/s3fs/core.py#L276)

src/litdata/streaming/downloader.py
```diff
@@ -188,10 +194,42 @@ def download_file(self, remote_filepath: str, local_filepath: str) -> None:
...
+        storage_options: Dict | None = {},
+    ):
+        remote_dir = remote_dir.replace("local:", "")
+        super().__init__(cloud_provider, remote_dir, cache_dir, chunks, storage_options)
+
+    def download_file(self, remote_filepath: str, local_filepath: str) -> None:
+        if os.path.exists(local_filepath) or remote_filepath == local_filepath:
+            return
+        try:
+            with FileLock(local_filepath + ".lock", timeout=3):
```
bhimrazy, line 222: Hi @deependujha, 

I was reviewing the documentation and came across the section on transactions. I’m wondering if this approach might be a good alternative to locking. 

Would love to hear your thoughts: https://filesystem-spec.readthedocs.io/en/latest/features.html#transactions

tchaton: We can combine both. The main challenge is to avoid multiple processes to override the same chunks. This is why we have the lock.

tchaton: Another option would be store the data in the RAM and write to disk once used.

deependujha:
tested successfully on S3 and GS for optimize(mode= none | append | overwrite), checkpoint, merge_datasets, streaming_dataset.

All the `# todo: add support for other providers` removed.

requirements/extras.txt
```diff
@@ -5,3 +5,6 @@ pyarrow
 tqdm
 lightning-cloud == 0.5.70 # Must be pinned to ensure compatibility
 google-cloud-storage
+fsspec[s3] # aws s3
```
bhimrazy, line 8: imo, I think it would be great if we could make at least one of them a default install option. (eg: s3)

tchaton: Yes, let's have s3 as the default. As it is the default supported on the Lightning Platform.

deependujha:
Tested on S3, GCP, Azure.

A repository for quick testing: https://github.com/deependujha/litdata-minimal-example-to-test/tree/main

### Remaining work

- Remove unnecessary `print` statements
- Decide if S3Client should be left commented, or the file to be deleted.

Anyways ready for review.

requirements/extras.txt
deependujha: done!

src/litdata/streaming/client.py
```diff
@@ -1,70 +1,70 @@
...
+#             )
+#         else:
+#           provider = InstanceMetadataProvider(iam_role_fetcher=InstanceMetadataFetcher(timeout=3600, num_attempts=5))
+#             credentials = provider.load()
+#             self._client = boto3.client(
+#                 "s3",
+#                 aws_access_key_id=credentials.access_key,
+#                 aws_secret_access_key=credentials.secret_key,
+#                 aws_session_token=credentials.token,
+#                 config=botocore.config.Config(retries={"max_attempts": 1000, "mode": "adaptive"}),
```
tchaton, line 56: Can we make this we have those defaults for S3 client and it is additive to whather the user adds as extra. 

```
default_storage_kwargs = {retries=retries={"max_attempts": 1000, "mode": "adaptive"}}
storage_options = {**default_storage_kwargs, storage_options}
```

tchaton (commented):
Most likely, you can delete the S3client things if the benchmark are conclusive.

src/litdata/processing/utilities.py
```diff
@@ -201,27 +199,26 @@ def read_index_file_content(output_dir: Dir) -> Optional[Dict[str, Any]]:
...
 
-        if obj.scheme != "s3":
-            raise ValueError(f"The provided folder should start with s3://. Found {output_dir.path}.")
-
-        # TODO: Add support for all cloud providers
-        s3 = boto3.client("s3")
-
-        prefix = obj.path.lstrip("/").rstrip("/") + "/"
+        if obj.scheme not in _SUPPORTED_CLOUD_PROVIDERS:
+            raise ValueError(
```
bhimrazy, line 203: Looking good so far. Just a small note: we should also add a test to ensure that this error is triggered when unsupported cloud provider paths are passed.

Does that work for you?

src/litdata/streaming/resolver.py
```diff
@@ -218,28 +219,27 @@ def _assert_dir_is_empty(output_dir: Dir, append: bool = False, overwrite: bool
 
     obj = parse.urlparse(output_dir.url)
 
-    if obj.scheme != "s3":
-        raise ValueError(f"The provided folder should start with s3://. Found {output_dir.url}.")
-
-    s3 = boto3.client("s3")
+    if obj.scheme not in _SUPPORTED_CLOUD_PROVIDERS:
```
bhimrazy, line 222: Looking good so far. Just a small note: we should also add a test to ensure that this error is triggered when unsupported cloud provider paths are passed.

I think currently some other error is taking precedence , may be I think if we could check in the `_resolve_dir` as well

src/litdata/streaming/client.py
deependujha:
```python
_DEFAULT_STORAGE_OPTIONS = {
    "s3":{"config_kwargs": {"retries":{"max_attempts": 1000, "mode": "adaptive"}}},
}

def get_complete_storage_options(cloud_provider: str, storage_options: Optional[Dict] = {}) -> Dict:
    if storage_options is None:
        storage_options = {}
    if cloud_provider in _DEFAULT_STORAGE_OPTIONS:
        return {**_DEFAULT_STORAGE_OPTIONS[cloud_provider], **storage_options}
    return storage_options
```

- Only adding for S3 now. `gcsfs` doesn't seem to have such retry storage_option. But, if it has we can extend from here on in future.

src/litdata/streaming/resolver.py
deependujha: interesting. For an unsupported cloud, it will use `return Dir(path=dir_path_absolute, url=None)`

Let's say if user asks for `dropbox://some_path`, resolve_dir will consider it to be a local_path and not cloud.

This was the original behaviour. Should we do something or leave it as it is? What do you think!

bhimrazy: I think we can check if the obj scheme is provided in URLs with supported providers.

deependujha: Done!

src/litdata/streaming/client.py
Borda, line 1: if all is commented, let's delete it?

deependujha: Sure. We were waiting for Benchmarking Litdata (this PR). But, I'll clean up the PR.

bhimrazy:
Here's a sneak peek of my version of the [benchmark](https://lightning.ai/lightning-ai/studios/benchmark-cloud-data-loading-libraries) tests. I think we might need to run a few more to ensure everything is in good shape. 

Image

cc: @Borda @tchaton

src/litdata/streaming/client.py
bhimrazy: Hi @deependujha , I think now, we are in somewhere good spot to start the cleanup and also updating the docs to use the fsspec version.

deependujha: sure.

deependujha:
For error

```txt
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
boto3 1.35.21 requires botocore<1.36.0,>=1.35.21, but you have botocore 1.35.16 which is incompatible.
```

- Run command

```bash
pip install -U boto boto3 botocore aiobotocore
```

tchaton (approved):
Awesome work !

PR #378, opened 2024-09-18
Bump version to 0.26.dev1

rasbt:
Bumps the version so it will be easier for users to differentiate between the latest stable release and the current dev version via the main branch when reporting LitData issues.

### DIFF
--- a/src/litdata/__about__.py
+++ b/src/litdata/__about__.py
@@ -14,7 +14,7 @@
 
 import time
 
-__version__ = "0.2.26"
+__version__ = "0.2.26.dev1"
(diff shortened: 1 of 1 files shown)


deependujha (approved):

src/litdata/__about__.py
```diff
@@ -14,7 +14,7 @@
 
 import time
 
-__version__ = "0.2.26"
+__version__ = "0.2.26.dev1"
```
Borda, line 17:
```suggestion
__version__ = "0.2.27.dev"
```

Borda (approved):

PR #381, opened 2024-09-22
Fix/large num chunks error

### LINKED ISSUE #366: Large number of chunks causes `OSError: [Errno 24] Too many open files`
## 🐛 Bug

If an optimized dataset has too many chunks (can be replicated using small chunk size or reasonable chunk size with lots of data), while reading (e.g. using `StreamingDataLoader`), a `Too many open files` error is encountered at some point. This is because each chunk is mmap'ed as it is loaded, and the mmap'ed handle is never released, leading to too many open files across the system:

https://github.com/Lightning-AI/litdata/blob/7efd76197e0191c0a07c9f0583ce9aec4b80a49e/src/litdata/streaming/item_loader.py#L320-L322

One could ofcourse use `ulimit` to increase it, but at some point the number of chunks can really be too high. 

It is easily resolved by keeping the number of mmap'ed files under a certain limit (for example by occasionally deleting some entries in the above dict, either randomly or smartly using FIFO or so).

I wanted to understand if this fix is alright, I am happy to send a PR for this if it makes sense. 

### To Reproduce

1. Optimize a dataset that results in a large number of chunks
2. Use StreamingDataset and StreamingDataLoader to load the data
3. At some point, `OSError: [Errno 24] Too many open files` will be encountered

Minimal code sample below to reproduce the issue.

#### Code sample
```python
import glob
import numpy as np
import random

from pathlib import Path

from litdata import optimize
from litdata.streaming import StreamingDataLoader, StreamingDataset, TokensLoader

# Fake tokenizer
def tokenize_fn(filepath):
    yield np.array([random.randint(0, 10000) for _ in range(random.randint(100, 1000))])

def main():
    Path("fake_file.txt").touch()
    outputs = optimize(
        fn=tokenize_fn,
        inputs=["fake_file.txt" for i in range(10000)], # increase number of files if error is not encountered on a specific machine
        output_dir="./optimized/",
        chunk_bytes="10KB",
        num_workers=1,
        item_loader=TokensLoader(block_size=1024),
    )
    


    train_dataset = StreamingDataset(
        input_dir="./optimized/",
        item_loader=TokensLoader(block_size=1024),
        shuffle=True,
        drop_last=False,
    )

    train_dataloader = StreamingDataLoader(
        train_dataset, batch_size=1, pin_memory=False, num_workers=1, drop_last=False
    )

    total_tokens = 0
    total_batches = 0
    for sample in train_dataloader:
        total_batches += 1
        total_tokens += np.prod(sample.shape)

        print(total_batches, total_tokens)
            
    print("Batches:", total_batches)
    print("Tokens:", total_tokens)

if __name__ == "__main__":
    main()
```

### Environment

- LitData version: 0.2.26
- OS (e.g., Linux): Linux
- How you installed: pip
- Python version: 3.12

bhimrazy:
<details>

</details>

## What does this PR do?

Fixes #366.

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/item_loader.py
+++ b/src/litdata/streaming/item_loader.py
@@ -363,6 +363,7 @@ class TokensLoader(BaseItemLoader):
     def close(self, chunk_index: int) -> None:
         """Release the memory-mapped file for a specific chunk index."""
         if chunk_index in self._mmaps:
+            self._mmaps[chunk_index]._mmap.close()
             del self._mmaps[chunk_index]
         if chunk_index in self._buffers:
             del self._buffers[chunk_index]
--- a/src/litdata/streaming/item_loader.py
+++ b/src/litdata/streaming/item_loader.py
@@ -363,9 +363,10 @@ class TokensLoader(BaseItemLoader):
     def close(self, chunk_index: int) -> None:
         """Release the memory-mapped file for a specific chunk index."""
         if chunk_index in self._mmaps:
-            del self._mmaps[chunk_index]  # Remove the memmap object
+            self._mmaps[chunk_index]._mmap.close()
+            del self._mmaps[chunk_index]
         if chunk_index in self._buffers:
-            del self._buffers[chunk_index]  # Remove the associated buffer
+            del self._buffers[chunk_index]
 
     @classmethod
     def encode_data(cls, data: List[bytes], _: List[int], flattened: List[Any]) -> Tuple[bytes, Optional[int]]:
--- a/src/litdata/streaming/item_loader.py
+++ b/src/litdata/streaming/item_loader.py
@@ -313,7 +313,7 @@ class TokensLoader(BaseItemLoader):
         offset = (1 + chunk["chunk_size"] + 1) * 4
         mmap = np.memmap(chunk_filepath, mode="r", order="C", offset=offset)
         self._mmaps[chunk_index] = mmap
-        self._buffers[chunk_index] = memoryview(mmap)  # type: ignore
+        self._buffers[chunk_index] = memoryview(mmap)  
 
     def pre_load_chunk(self, chunk_index: int, chunk_filepath: str) -> None:
         # This is called within the prepare chunks thread, so we overlap data loading with data reading.
--- a/src/litdata/streaming/item_loader.py
+++ b/src/litdata/streaming/item_loader.py
@@ -360,6 +360,13 @@ class TokensLoader(BaseItemLoader):
                 del self._mmaps[chunk_index]
             os.remove(chunk_filepath)
 
+    def close(self, chunk_index: int) -> None:
+        """Release the memory-mapped file for a specific chunk index."""
+        if chunk_index in self._mmaps:
+            del self._mmaps[chunk_index]  # Remove the memmap object
+        if chunk_index in self._buffers:
+            del self._buffers[chunk_index]  # Remove the associated buffer
+
     @classmethod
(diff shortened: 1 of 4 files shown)


deependujha (approved):

src/litdata/streaming/item_loader.py
```diff
@@ -360,6 +360,13 @@ def delete(self, chunk_index: int, chunk_filepath: str) -> None:
                 del self._mmaps[chunk_index]
             os.remove(chunk_filepath)
 
+    def close(self, chunk_index: int) -> None:
+        """Release the memory-mapped file for a specific chunk index."""
+        if chunk_index in self._mmaps:
```
tchaton, line 365: We need to close the self._mmaps before deleting it. However, this used to cause some failures.

bhimrazy: It seems `np.memmap` does not provide a direct `close` method.

> [Currently, there is no API to close the underlying mmap. It is tricky to ensure the resource is actually closed, since it may be shared between different memmap instances.](https://numpy.org/doc/stable/reference/generated/numpy.memmap.html#numpy.memmap)

However, in this issue (https://github.com/numpy/numpy/issues/13510), I found a way to close it with the following:

```python
self._mmaps[chunk_index]._mmap.close()
```

I'll try to implement this and check for any potential errors.

fdalvi:
Hi @bhimrazy,

Thanks a lot for taking this up! Work got really busy so I didn't get to implement a potential fix.

I took a look at the current changes, I am not 100% confident I understand the changes fully, but it seems like we will close the mmap file after every item read (and then open it on every successive read because its not in the cache)? This is a bit excessive in terms of going the way of almost not caching anything right?

Please let me know if I have misunderstood the implementation. Thanks again!

Best,

src/litdata/streaming/reader.py
```diff
@@ -305,6 +304,9 @@ def read(self, index: ChunkedIndex) -> Any:
             # track the new chunk index as the latest one
             self._last_chunk_index = index.chunk_index
 
+        if isinstance(self._item_loader, TokensLoader):
```
tchaton, line 307: This needs to be under the previous if, so it is called when the entire chunk has been consumed, not on every read call ;)

bhimrazy: Thanks for pointing that out, @tchaton! I’ll take a closer look and make the necessary adjustments.

bhimrazy:
> I took a look at the current changes, I am not 100% confident I understand the changes fully, but it seems like we will close the mmap file after every item read (and then open it on every successive read because its not in the cache)? This is a bit excessive in terms of going the way of almost not caching anything right?

Yes, @fdalvi, you’re correct. Thanks for bringing this up! I’ll review the changes in more detail and make the necessary adjustments shortly.

src/litdata/streaming/reader.py
```diff
@@ -302,6 +301,11 @@ def read(self, index: ChunkedIndex) -> Any:
             # inform the chunk has been completely consumed
             self._prepare_thread.delete([self._last_chunk_index])
 
+        if index.chunk_index != self._last_chunk_index:
+            # Close the memory-mapped file for the last chunk index
+            if isinstance(self._item_loader, TokensLoader) and self._last_chunk_index is not None:
+                self._item_loader.close(self._last_chunk_index)
+
```
bhimrazy, line 308: Moved it to a separate condition to run independently of `(self._config._remote_dir or self._config._compressor)`

tchaton (approved):

PR #415, opened 2024-11-13
Change S3Client to use user-provided storage_options even in Studio 

### LINKED ISSUE #414: use storage_options even when IS_IN_STUDIO
## 🚀 Feature

Can the streaming.client.S3Client be modified to use the passes storage_options when they are passed in, even when in a STUDIO? Currently I pass in storage_options and they are ignored and STUDIO credentials are used instead. I would like to be able to use my storage credentials, but there's currently no easy way to do that.

### Motivation

It's confusing when passed storage_options aren't used, difficult to figure out why private objects cannot be accessed when the passed storage_options should provide sufficient permissions. I know you can connect a studio to an s3 bucket, but I shouldn't have to if I've passed the required storage_options.

### Pitch

Modify the S3Client code from this:
```
if has_shared_credentials_file or not _IS_IN_STUDIO:
```

to this:
```
if has_shared_credentials_file or not _IS_IN_STUDIO or self._storage_options:
```

### Alternatives

It seems like setting shared_credentials_file is one way, but even still, when I pass storage_options, I would expect them to be used.

### Additional context

NA

grez72:
## What does this PR do?

This PR fixes an issue in the S3Client class where user-provided storage_options are ignored in the Lightning AI Studio environment. The client now uses user-supplied credentials when provided in storage_options.

Fixes #414

### DIFF
--- a/src/litdata/streaming/client.py
+++ b/src/litdata/streaming/client.py
@@ -37,7 +37,7 @@ class S3Client:
             os.getenv("AWS_SHARED_CREDENTIALS_FILE") == os.getenv("AWS_CONFIG_FILE") == "/.credentials/.aws_credentials"
         )
 
-        if has_shared_credentials_file or not _IS_IN_STUDIO:
+        if has_shared_credentials_file or not _IS_IN_STUDIO or self._storage_options:
             self._client = boto3.client(
                 "s3",
                 **{


src/litdata/streaming/client.py
```diff
@@ -37,7 +37,7 @@ def _create_client(self) -> None:
             os.getenv("AWS_SHARED_CREDENTIALS_FILE") == os.getenv("AWS_CONFIG_FILE") == "/.credentials/.aws_credentials"
         )
 
-        if has_shared_credentials_file or not _IS_IN_STUDIO:
+        if has_shared_credentials_file or not _IS_IN_STUDIO or self._storage_options:
```
tchaton, line 40: It might be better to propagate the storage options to the client below. I found the provider one to be more reliable in general.

bhimrazy:
Hi @grez72,

Would you mind enabling "Allow edits by maintainers"? This will help us assist with any necessary updates directly. 
Please let us know if there’s anything further we can help with. 

Thank you!

bhimrazy:
Thanks for clarifying, @grez72!

In that case, could you please assist with the final updates based on @tchaton's suggestion? Here’s a proposed change that should help integrate `self._storage_options` into the client setup:

```diff
-      if has_shared_credentials_file or not _IS_IN_STUDIO or self._storage_options:
+      if has_shared_credentials_file or not _IS_IN_STUDIO
            self._client = boto3.client(
                "s3",
                **{
                    "config": botocore.config.Config(retries={"max_attempts": 1000, "mode": "adaptive"}),
                    **self._storage_options,
                },
            )
        else:
            provider = InstanceMetadataProvider(iam_role_fetcher=InstanceMetadataFetcher(timeout=3600, num_attempts=5))
            credentials = provider.load()
            self._client = boto3.client(
                "s3",
                aws_access_key_id=credentials.access_key,
                aws_secret_access_key=credentials.secret_key,
                aws_session_token=credentials.token,
                config=botocore.config.Config(retries={"max_attempts": 1000, "mode": "adaptive"}),
+              **self._storage_options,
            )
```

grez72:
I implemented the suggested change, but ran into a few issues when passing credentials as storage options like so:
```
dataset = StreamingDataset(
    "s3://path/to/streaming-dataset,
    storage_options=dict(aws_access_key_id=..., aws_secret_access_key=..., endpoint_url=...)
)
```

First I encountered this error (which makes sense, my storage options also contain aws_access_key_id).
```
boto3.client() got multiple values for keyword argument 'aws_access_key_id'
``` 

I next tried to merge the provider credentials and self._storage_options, like so:
```
provider = InstanceMetadataProvider(iam_role_fetcher=InstanceMetadataFetcher(timeout=3600, num_attempts=5))
            credentials = provider.load()            
            self._client = boto3.client(
                "s3",
                **{
                   "config": botocore.config.Config(retries={"max_attempts": 1000, "mode": "adaptive"}),
                    **dict(aws_access_key_id=credentials.access_key,
                          aws_secret_access_key=credentials.secret_key,
                          aws_session_token=credentials.token), 
                    **self._storage_options
                },
            )
```

But here I get a permissions error because provider.load() provides a `aws_session_token` that doesn't get overwritten by my storage_options (which only has the key_id and secret key). This could be avoided by passing aws_session_token=None with storage_credentials, but that seems kind of counterintutive from the user point of view.

I think maybe if the user passes storage options that include aws credentials (e.g., 'aws_access_key_id' or 'aws_session_token'), then none of the credentials returned by the provider are relevant, so maybe the following would be best? I'm not sure how the aws_session_token works (so maybe checking for that is not needed). If this seems reasonable I can make the change.

```
has_shared_credentials_file = (
            os.getenv("AWS_SHARED_CREDENTIALS_FILE") == os.getenv("AWS_CONFIG_FILE") == "/.credentials/.aws_credentials"
        )

        storage_options_inlude_credentials = (
            'aws_access_key_id' in self._storage_options or 'aws_session_token' in self._storage_options
        )
        
        if has_shared_credentials_file or not _IS_IN_STUDIO or (_IS_IN_STUDIO and storage_options_inlude_credentials):
            self._client = boto3.client(
                "s3",
                **{
                    "config": botocore.config.Config(retries={"max_attempts": 1000, "mode": "adaptive"}),
                    **self._storage_options,
                },
            )
        else:
            provider = InstanceMetadataProvider(iam_role_fetcher=InstanceMetadataFetcher(timeout=3600, num_attempts=5))
            credentials = provider.load()
            self._client = boto3.client(
                "s3",
                aws_access_key_id=credentials.access_key,
                aws_secret_access_key=credentials.secret_key,
                aws_session_token=credentials.token,
                config=botocore.config.Config(retries={"max_attempts": 1000, "mode": "adaptive"}),
            )
```

tchaton:
Yes, fine by me ;)

Issue #423, opened 2024-11-27
tif inferred as pickle


robmarkcole:
## 🐛 Bug

Returning paths to tifs, but they are inferred as pickle:

```python
# dataset yields paths 

test_dataset[0] == {'image_path': '/teamspace/studios/this_studio/dataset/test/T20MMT_20200505T142729_tile_17_11.tif',
 'mask_path': '/teamspace/studios/this_studio/dataset/test/T20MMT_20200505T142729_tile_17_11_mask.tif',
 'image_id': 'T20MMT_20200505T142729_tile_17_11'}
```
Logs
```
Rank 1 inferred the following `['str', 'pickle', 'pickle']` data format.
```
But expect:
```
Rank 1 inferred the following `['str', 'tif', 'tif']` data format. 
```

The script generates the same number of chunk files for both train and test, although train has significantly more tifs..

litdata==0.2.19

Reproducible example:
```python
from litdata.streaming.serializers import _SERIALIZERS

image_path = '/teamspace/studios/this_studio/dataset/test/T10VDK_20200519T193911_tile_0_7.tif'

def evaluate_serializer(filename: str) -> str:
    """
    Evaluate which serializer would be used for a given filename.

    Args:
        filename: The name of the file to evaluate.

    Returns:
        The name of the selected serializer.
    """
    # Iterate through serializers in the order defined in _SERIALIZERS
    for serializer_name, serializer in _SERIALIZERS.items():
        if serializer.can_serialize(filename):
            return serializer_name

    # If no serializer can handle the file, raise an exception
    raise ValueError(f"No suitable serializer found for filename: {filename}")

assert evaluate_serializer(image_path) == 'pickle'
```

robmarkcole:
OK,[ I see ](https://github.com/Lightning-AI/litdata/blob/7ba51acd194f5c34a0757ccbcf836fa68d5d743f/src/litdata/streaming/serializers.py#L316)
```
# FileSerializer will be removed in the future.
```

PR #425, opened 2024-11-28
POC: add tiffile serializer

robmarkcole:
## What does this PR do?

Fixes [# (issue).](https://github.com/Lightning-AI/litdata/issues/423)

## PR review

## Did you have fun?


tchaton (commented):
Can you add a test ?

src/litdata/streaming/serializers.py
```diff
@@ -387,13 +388,37 @@ def can_serialize(self, data: float) -> bool:
         return isinstance(data, float)
 
 
+class TIFFSerializer(Serializer):
```
tchaton, line 391: > Many compression and predictor schemes are supported via the imagecodecs library, including LZW, PackBits, Deflate, PIXTIFF, LZMA, LERC, Zstd, JPEG (8 and 12-bit, lossless), JPEG 2000, JPEG XR, JPEG XL, WebP, PNG, EER, Jetraw, 24-bit floating-point, and horizontal differencing.

From the tifffile, it seems it comes with compression built-in. It might be worth to see if there is a compressed version of the file to use

tchaton: Also, this is the kind of serializer you want to customize and pass to the `optimize` and `StreamingDataset` to be max performance from it.

robmarkcole: Would be interesting to explore that as an option - however we generally don't want to compress remote sensing imagery.

tchaton: From checking the docs, users are free to apply any compression while generating tiff files using imwrite. So I guess it is fine.

src/litdata/streaming/serializers.py
```diff
@@ -387,13 +388,36 @@ def can_serialize(self, data: float) -> bool:
...
+            raise ValueError(f"The item to serialize must be a valid file path. Received: {item}")
+
+        # Read the TIFF file as bytes
+        with open(item, "rb") as f:
+            data = f.read()
+
+        # Store metadata if needed
+        _, ext = os.path.splitext(item)
+        metadata = f"tifffile:{ext.lower()}"
+        return data, metadata
```
tchaton, line 405: You don t need metadata. It is used onlt when restoring data conditional like dtype for tensors.

robmarkcole: Seems necessary to return None

tchaton (approved):
LGTM !

PR #438, opened 2024-12-22
Feat: Add `per_stream` batching method to CombinedStreamingDataset

### LINKED ISSUE #434: Advanced Batching Logic with CombinedStreamingDataset
## 🚀 Feature

When we use `CombinedStreamingDataset`, samples are drawn across all datasets within a single batch. Other packages like MosaicML's StreamingDataset allows you to [specify advanced strategies for batching](https://docs.mosaicml.com/projects/streaming/en/latest/dataset_configuration/mixing_data_sources.html#batching-methods).

The current implementation maps to the `random` option from MosaicML. It would great if we could support other batch methods -- especially `Per Stream`, where in each batch is taken from a single dataset. 

### Motivation

If you're training over K datasets, where each dataset yields samples of different shapes - we currently have to split the `random batch` into microbatches, since the samples from the different datasets cannot be stacked together. These leads to 10-20% slowdown in training, because we're not fully utilizing the GPU on ever forward and backward pass. 

### Pitch

Allow for additional options in batching, e.g. "per stream" -- which dictates how batches are yielded, when using `CombinedStreamingDataset`.

schopra8:
<details>

</details>

### What does this PR do?

Adds support for `batching_method` :`per_stream` batching to `CombinedStreamingDataset`, addressing #434.  
This new batching strategy ensures that each batch is sampled exclusively from a single dataset, which is helpful when datasets have varying shapes or structures.

Fixes #434 

## PR review

## Did you have fun?


schopra8:
@tchaton When is the `CacheDataLoader` used vs `StreamingDataLoader` in conjunction with a `CombinedStreamingDataset`?

I'm trying to send messages to iterators and process them within their worker loops -- when the iterator must change it's dataset index (i.e., on batch boundaries). I'm getting confused by the difference between  `_MultiProcessingDataLoaderIterPatch` and `_StreamingMultiProcessingDataLoaderIter`. Seems like the former is used in `CacheDataLoader` while the latter is used in `StreamingDataLoader`.

src/litdata/streaming/dataloader.py
```diff
@@ -626,17 +669,18 @@ def __iter__(self) -> Any:
                 self._num_samples_yielded_streaming += self.batch_size
                 yield batch
         else:
+            # Assume, this is a CombinedStreamingDataset.
             self.dataset._set_use_streaming_dataloader(True)
             assert self.batch_size
-            # TODO: Inject a custom collate function to avoid collating the __NUM_SAMPLES_YIELDED__ key
+
             for batch in super().__iter__():
+                print("Fetched a batch ...")
```
Borda, line 677: maybe use loggier instead of print

Borda (approved):

bhimrazy:
Thank you @schopra8 for the PR and your patience — we'll start looking into this shortly and aim to land it soon.

src/litdata/streaming/combined.py
```diff
@@ -127,6 +128,14 @@ def set_batch_size(self, batch_size: int) -> None:
         for dataset in self._datasets:
             dataset.set_batch_size(batch_size)
 
+    def set_batching_method(self, batching_method: Literal["stratified", "per_stream"]) -> None:
+        """Set the current batching method to the datasets.
+        When batching_method is "stratified" (default), batches consist of samples from all datasets.
+        When batching_method is "per_stream" batches consist of samples from one dataset,
+        which is selected at random.
+        """
+        self.batching_method = batching_method
```
Borda, line 137: in such case this shall be
```suggestion
        self._batching_method = batching_method
```
or make it a property so it can'o't be randomly edited...

src/litdata/streaming/dataloader.py
```diff
@@ -476,6 +495,21 @@ def __init__(self, loader: DataLoader) -> None:
...
+        if (
+            isinstance(self._loader.dataset, CombinedStreamingDataset)
+            and self._loader.dataset.batching_method == "per_stream"
+            and self._rcvd_idx % self._loader.batch_size == 0
+        ):
+            print(f"Batch {self._rcvd_idx // self._loader.batch_size}: Sending SET_NEW_DATASET_INDEX command to worker")
+            self._index_queues[self._loader._latest_worker_idx].put(("SET_NEW_DATASET_INDEX", None))
+
+        return data
+
```
bhimrazy, line 512: looking into better way to set the dataset index as letting the DataLoader directly control the dataset index selection for "per_stream" batching is not ideal.

bhimrazy: Idea:

1. The iterator `[_CombinedDatasetIterator]` should be responsible for both "stratified" and "per_stream" batching.
2. The DataLoader should only ask for the next batch; it should not know or care how the batches are constructed.

src/litdata/streaming/combined.py
```diff
@@ -272,6 +292,26 @@ def __next__(self) -> Any:
         return self._get_sample(self._get_dataset_index())
 
     def _get_dataset_index(self) -> int:
+        if self._batching_method == BatchingMethod.STRATIFIED:
+            # For every sample, randomly select a dataset (weighted)
+            dataset_idx = self._set_new_dataset_index()
+        elif self._batching_method == BatchingMethod.PER_STREAM:
+            # For each batch, pick a dataset and stick with it for the whole batch
+            # You need to track how many samples have been yielded in the current batch
+            if not hasattr(self, "_samples_yielded_in_batch"):
```
tchaton, line 301: You can set those in the __init__ of the class

tchaton (approved):
Nice @bhimrazy !

PR #468, opened 2025-02-12
Using count-locks for multi-node-single-cache support

JackUrb:
<details>
  <summary><b>Before submitting</b></summary>

- [x] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lit-data/blob/main/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes #408.

The basic approach is to be _much_ more strict at tracking the current number of readers that expect a specific file to exist, incrementing a lock when these are downloaded (or requested), and decrementing the lock whenever a reader is "done" with a shard.

At the moment it _does_ lead to stable multi-node runs that share the same cache, however mostly due to the fact that the majority of files are never allowed to delete, as there are more lock calls than unlock calls. Debugging this now. This suggests that it is the single lock limitation that is causing the underlying multi-node issue, and getting the rest of this correct should resolve it overall.

This PR doesn't address the unzipping part, which likely just needs a timeout.

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in GitHub issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #471, opened 2025-02-14
fix: _get_folder_size fn

deependujha:
<details>

</details>

## What does this PR do?

This PR addresses comment: https://github.com/Lightning-AI/litdata/pull/468#issuecomment-2659248600

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/reader.py
+++ b/src/litdata/streaming/reader.py
@@ -439,10 +439,13 @@ def _get_folder_size(path: str) -> int:
 
     """
     size = 0
+    _SUPPORTED_EXTENSIONS = (".bin", ".json", ".parquet", ".npy")
+
     for dirpath, _, filenames in os.walk(str(path)):
         for filename in filenames:
-            with contextlib.suppress(FileNotFoundError):
-                size += os.stat(os.path.join(dirpath, filename)).st_size
+            if filename.endswith(_SUPPORTED_EXTENSIONS):
+                with contextlib.suppress(FileNotFoundError):
+                    size += os.stat(os.path.join(dirpath, filename)).st_size
     return size
 
 


src/litdata/streaming/reader.py
```diff
@@ -439,10 +439,13 @@ def _get_folder_size(path: str) -> int:
...
     size = 0
+    _SUPPORTED_EXTENSIONS = (".bin", ".json", ".parquet", ".npy")
+
     for dirpath, _, filenames in os.walk(str(path)):
         for filename in filenames:
-            with contextlib.suppress(FileNotFoundError):
-                size += os.stat(os.path.join(dirpath, filename)).st_size
+            if filename.endswith(_SUPPORTED_EXTENSIONS):
+                with contextlib.suppress(FileNotFoundError):
+                    size += os.stat(os.path.join(dirpath, filename)).st_size
```
tchaton, line 448: The idea is not use os.stat which is expensive and not walk the folder.

Instead, we check only the cache folder, not its parents, list the file and use the config to estimate the size

tchaton (commented):
Note

deependujha:
sorry that was just to make the pr, I was working on using config dict.

src/litdata/streaming/reader.py
```diff
@@ -182,7 +182,10 @@ def _maybe_delete_chunks(self) -> None:
     def _can_delete_chunk(self) -> bool:
         if self._delete_chunks_when_processed:
             return self._pre_download_counter >= self._max_pre_download - 1
-        return self._max_cache_size is not None and _get_folder_size(self._parent_cache_dir) >= self._max_cache_size
+        return (
+            self._max_cache_size is not None
+            and _get_folder_size(self._parent_cache_dir, self._config) >= self._max_cache_size
```
tchaton, line 187: We shouldn't use the `parent_dir` anymore. Otherwise, this would always be empty. Each StreamingDataset should take care only its own cache.

src/litdata/streaming/reader.py
bhimrazy: > Instead, we check only the cache folder, not its parents, list the file and use the config to estimate the size

yes, while debugging to understand the sizes, I could see it was considering the other dirs(for eg: `remote_dir`) in the tests

bhimrazy:
@deependujha, I believe the failing tests are correct because, previously, the size calculation included the `remote_dir`, whereas now it only considers the `cachedir` size. Since the `max_cache_size` remains unchanged, this results in more chunks being stored in the cache directory.  

To address this, I think we could either reduce the `max_cache_size` to around 1300 or update the number of files we check.

src/litdata/streaming/reader.py
```diff
@@ -432,17 +435,17 @@ def __del__(self) -> None:
...
+def _get_folder_size(path: str, config: ChunksConfig) -> int:
     """Collect the size of each files within a folder.
 
     This method is robust to file deletion races
 
     """
     size = 0
-    for dirpath, _, filenames in os.walk(str(path)):
-        for filename in filenames:
+    for filename in os.listdir(os.path.join(path, "cache_dir")):
```
tchaton, line 445: We shouldn't add cache_dir there.

src/litdata/streaming/reader.py
```diff
@@ -432,17 +435,17 @@ def __del__(self) -> None:
...
+        if filename in config.filename_to_size_map:
             with contextlib.suppress(FileNotFoundError):
-                size += os.stat(os.path.join(dirpath, filename)).st_size
+                size += config.filename_to_size_map[filename]
```
tchaton, line 448: Let's check if filename is within the map, otherwise, this would fail the thread. Normally, the files should be. We need to print a warning if it doesn't.

src/litdata/streaming/reader.py
bhimrazy: I think we can use `self._config._cache_dir` here.

src/litdata/streaming/reader.py
bhimrazy: Config does seem to have the `cache_dir`, so may be we could use from there.`self._config._cache_dir`

src/litdata/streaming/reader.py
deependujha: the code is already in `if` block:

```python
def _get_folder_size(path: str, config: ChunksConfig) -> int:
    """Collect the size of each files within a folder.

    This method is robust to file deletion races

    """
    size = 0
    for filename in os.listdir(path):
        if filename in config.filename_to_size_map: # only if filename is in config map
            with contextlib.suppress(FileNotFoundError):
                size += config.filename_to_size_map[filename]
    return size
```

src/litdata/streaming/reader.py
deependujha: yeah, I forgot to update that. too many context-switching 🥲

src/litdata/streaming/reader.py
bhimrazy: Looks good! Just missing a warning if the filename isn’t found in the map.

deependujha: cnt and lock files won't be found in the map. So this is expected.

tchaton: Yes, we should skip those in the listing.

deependujha: in that case, PR is ready to be merged. 🫣

deependujha: oh sorry, got it. If file is not `.cnt` or `.lock` and is still not found in the config map, then warn user.

tchaton (approved):
Looks great !

PR #478, opened 2025-02-19
🛠️ Fix: Ensure `chunk_bytes` in `index.json` matches actual chunk file size

### LINKED ISSUE #477: Mismatch Between index.json and Actual Chunk File Sizes
## 🐛 Bug Report: Mismatch Between index.json and Actual Chunk File Sizes

There seems to be discrepancy between the chunk sizes listed in index.json and the actual sizes of the corresponding chunk files stored on disk. 

### To Reproduce

1. Create a sample [optimized dataset](https://github.com/Lightning-AI/litdata?tab=readme-ov-file#speed-up-model-training)
2. Check the size of chunk files and compare it with present in the index.json file

Image
```
ls -l fast_data/
```

### Expected behavior
The actual file sizes should match the chunk_bytes values recorded in index.json.

bhimrazy:
## What does this PR do?
This PR fixes the inconsistency between the `chunk_bytes` value in index.json and the actual file size of chunk files.

Fixes #477.

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/writer.py
+++ b/src/litdata/streaming/writer.py
@@ -235,12 +235,11 @@ class BinaryWriter:
         sample_data = b"".join([item.data for item in items])
         data = num_items.tobytes() + offsets.tobytes() + sample_data
 
-        current_chunk_bytes = sum([item.bytes for item in items])
-
         # Whether to encrypt the data at the chunk level
         if self._encryption and self._encryption.level == EncryptionLevel.CHUNK:
             data = self._encryption.encrypt(data)
-            current_chunk_bytes = len(data)
+
+        current_chunk_bytes = len(data)
 
         if self._chunk_bytes and current_chunk_bytes > self._chunk_bytes:
             warnings.warn(
--- a/tests/streaming/test_writer.py
+++ b/tests/streaming/test_writer.py
@@ -251,6 +251,27 @@ def test_writer_unordered_indexes(tmpdir):
     assert data["chunks"][2]["chunk_size"] == 2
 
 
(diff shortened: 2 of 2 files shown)


src/litdata/streaming/config.py
```diff
@@ -237,8 +237,9 @@ def __getitem__(self, index: ChunkedIndex) -> Tuple[str, int, int]:
 
         filesize_bytes = chunk["chunk_bytes"]
 
-        if self._config and self._config.get("encryption") is None and (not local_chunkpath.endswith(".parquet")):
-            filesize_bytes += (1 + chunk["chunk_size"]) * 4
+        # TODO: Check why filesize_bytes is calculated this way.
+        # if self._config and self._config.get("encryption") is None and (not local_chunkpath.endswith(".parquet")):
+        #     filesize_bytes += (1 + chunk["chunk_size"]) * 4
```
bhimrazy, line 242: I still have abit of doubt if this could have any backward compatibilty issue. 
Will do some more tests on this.

https://github.com/Lightning-AI/litdata/blob/012baa28d7707fefa0bc47c1cb8ecfc054fd352f/src/litdata/streaming/item_loader.py#L161-L164

bhimrazy: After debugging and analysis, it appears that there won't be any issues. Previously, `filesize_bytes` was smaller than the actual file size, but after the update, it will match the actual size. Therefore, the condition `os.stat(chunk_filepath).st_size >= filesize_bytes` would always be true  for the particular chunk file.

tchaton (approved):
Real nice !

PR #486, opened 2025-02-27
Prototype: Moving Some Operations to Rust (PyO3) for Performance Testing

deependujha:
<details>

</details>

## What does this PR do?

Adds experimental rust support and aims to improve performance.

‼️ Warning:

- Not updating `github action` to publish to pypi for now. If we decide to merge this PR, it will be added.

## PR review

## Did you have fun?


src/rust_impl/fs/s3.rs
```diff
@@ -0,0 +1,302 @@
...
+            .splitn(2, '/')
+            .collect();
+        if parts.len() < 2 {
+            return Err("Invalid S3 path format. Expected 's3://bucket-name/key'".into());
+        }
+        let bucket_name = parts[0];
+        let key = parts[1];
+        // println!("in byte-range download fn with: bucket_name: {bucket_name} & key: {key}");
+        // Step 1: Get file size
+        let head_response = self
```
tchaton, line 240: We don't need the head request because we know the file length.

src/rust_impl/fs/s3.rs
```diff
@@ -0,0 +1,302 @@
...
+                let range_header = format!("bytes={}-{}", start, end);
+                let response = s3client
+                    .get_object()
+                    .bucket(bucket_name)
+                    .key(key)
+                    .range(range_header)
+                    .send()
+                    .await?;
+
+                let body = response.body.collect().await?;
```
tchaton, line 285: Here, we could just keep that in memory for now and have another method to let python access it.

src/rust_impl/fs/s3.rs
```diff
@@ -0,0 +1,302 @@
...
+            let key = key.to_string();
+            let file = Arc::clone(&file);
+
+            tasks.spawn(async move {
+                let range_header = format!("bytes={}-{}", start, end);
+                let response = s3client
+                    .get_object()
+                    .bucket(bucket_name)
+                    .key(key)
+                    .range(range_header)
```
bhimrazy, line 281: Oh wow, so the sdk already supports the byte ranges

src/litdata/streaming/config.py
```diff
@@ -109,6 +128,10 @@ def can_delete(self, chunk_index: int) -> bool:
             return True
         return chunk_index not in self._skip_chunk_indexes_deletion
 
+    def set_epoch(self, epoch: int) -> None:
+        self._epoch = epoch
+        self.streaming_data_provider.set_epoch(epoch)
```
tchaton, line 133: I am not sure why you need the epoch there.

src/litdata/streaming/cache.py
```diff
@@ -47,6 +48,10 @@ def __init__(
         writer_chunk_index: Optional[int] = None,
         storage_options: Optional[Dict] = {},
         max_pre_download: int = 2,
+        epoch: Optional[int] = None,
+        config: Optional[ChunksConfig] = None,
+        on_start_pre_item_download_count: int = 100,
```
tchaton, line 53: I think this was a good idea to try this out but I think we should just pass download_bytes_size="1MB"

src/litdata/streaming/cache.py
```diff
@@ -168,3 +180,8 @@ def _get_chunk_index_from_index(self, index: int) -> Tuple[int, int]:
     def save_checkpoint(self, checkpoint_dir: str = ".checkpoints") -> Optional[str]:
         """Save the current state of the writer to a checkpoint."""
         return self._writer.save_checkpoint(checkpoint_dir=checkpoint_dir)
+
+    def set_epoch(self, epoch: int) -> None:
+        """Set the epoch number."""
+        self._epoch = epoch
+        self._reader.set_epoch(epoch)
```
tchaton, line 187: I still don't fully understand why you need the epoch. The downloaders never needed it before.

src/litdata/streaming/reader.py
```diff
@@ -372,46 +430,77 @@ def read(self, index: ChunkedIndex) -> Any:
 
             if self._last_chunk_index is None:
                 self._last_chunk_index = index.chunk_index
+        if _USE_RUST_IMPLEMENTATION and self._streaming_chunks_downloader is None:
+            # print(f"{self.should_start_streaming_chunks_downloader}")
+            self._streaming_chunks_downloader = StreamingChunksDownloader(
```
tchaton, line 435: I am not sure why you need this.

src/litdata/streaming/config.py
deependujha: the initial idea was: rust streamDownloader will have two sets. one set contains the order for the current epoch, and another set for the upcoming epoch.

`current_downloader_epoch & current_streamer_epoch`. Downloader can be at max 1 epoch ahead. `config.py` updates epoch.

---

But, now many of these changes don't make sense much

src/litdata/streaming/cache.py
deependujha: bcoz downloader can start downloading initial indices for the next epoch. Else, after each epoch, we will again have performance down (something like cold start), where downloader will restart from start.

tchaton: Yes, but this works only if there are persistent workers with the DataLoader. But this isn't always the case.

PR #489, opened 2025-03-02
Warn user if `max_cache_size` is less than 25GB in StreamingDataset

bhimrazy:
## What does this PR do?
- Warn user about performance implications if max cache size is less than 25GB
- update deprecated `warn` method to `warning`
- Add `.tmp` to ignore files for cache size calculation

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/reader.py
+++ b/src/litdata/streaming/reader.py
@@ -451,7 +451,7 @@ def _get_folder_size(path: str, config: ChunksConfig) -> int:
         if filename in config.filename_to_size_map:
             with contextlib.suppress(FileNotFoundError):
                 size += config.filename_to_size_map[filename]
-        elif not filename.endswith((".cnt", ".lock", ".json", ".zstd.bin")):
+        elif not filename.endswith((".cnt", ".lock", ".json", ".zstd.bin", ".tmp")):
             # ignore .cnt, .lock, .json and .zstd files for warning
             logger.warning(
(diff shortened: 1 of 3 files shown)


src/litdata/streaming/reader.py
```diff
@@ -451,7 +451,7 @@ def _get_folder_size(path: str, config: ChunksConfig) -> int:
         if filename in config.filename_to_size_map:
             with contextlib.suppress(FileNotFoundError):
                 size += config.filename_to_size_map[filename]
-        elif not filename.endswith((".cnt", ".lock", ".json", ".zstd.bin")):
+        elif not filename.endswith((".cnt", ".lock", ".json", ".zstd.bin", ".tmp")):
```
deependujha, line 454: any reason for `.tmp`?

Also can you try optimizing a dataset and then streaming. In my case, logger was logging too many stuff (`chunk.bin.abcde`) and had to comment that.

bhimrazy: Sure. 

For the `.tmp`: I noticed the .tmp files appearing a few times when using a separate cache directory locally. It seems to be caused by LocalDownloader, likely due to the behavior in the following code:
https://github.com/Lightning-AI/litdata/blob/f6660be93a97346c4da9feb81011544b93d88c13/src/litdata/streaming/downloader.py#L194-L196

image

So, I thought of adding it as well.

Issue #495, opened 2025-03-05
Add register function to provide an "official" way to build custom downloader


ouj:
## 🚀 Feature

Allow extending litdata to support other storage protocols. 

### Motivation

There are storage solution like `dbfs` unsupported by litdata. Adding a way to register downloaders will make it easier to support different protocal.

### Pitch
This is an alternative to implementing dbfs downloader inside `litdata` https://github.com/Lightning-AI/litdata/issues/494

### Alternatives

https://github.com/Lightning-AI/litdata/issues/494

PR #496, opened 2025-03-05
Add register function for downloader

ouj:
<details>

</details>

## What does this PR do?

1. Add a method `register_downloader` to allow registering a new DOWNLOADER. If `overwrite=False`, it will raise error when a prefix is set twice.
2. Remove `LocalDownloader` from `_DOWNLOADERS` since "" will match any prefix. Practically made it the fallback of everything. Because `dict` is ordered, so it will stop the forloop in `get_downloader` before it can get to any newly registered downloader.
3. Add `unregiester_downloader` to unregister a downloader (for cleanup?)
4. Rename `get_downloader_cls` to `get_downloader` because the method return a downloader object not `type[Downloader]`.

Fixes https://github.com/Lightning-AI/litdata/issues/495

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/downloader.py
+++ b/src/litdata/streaming/downloader.py
@@ -249,6 +249,16 @@ _DOWNLOADERS: dict[str, type[Downloader]] = {
 
 
 def register_downloader(prefix: str, downloader_cls: type[Downloader], overwrite=False) -> None:
+    """Register a new downloader class with a specific prefix.
+
+    Args:
+        prefix (str): The prefix associated with the downloader.
+        downloader_cls (type[Downloader]): The downloader class to register.
+        overwrite (bool, optional): Whether to overwrite an existing downloader with the same prefix. Defaults to False.
+
+    Raises:
+        ValueError: If a downloader with the given prefix is already registered and overwrite is False.
+    """
     if prefix in _DOWNLOADERS and not overwrite:
         raise ValueError(f"Downloader with prefix {prefix} already registered.")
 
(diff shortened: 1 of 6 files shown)


src/litdata/streaming/downloader.py
```diff
@@ -239,20 +239,32 @@ def download_file(self, remote_filepath: str, local_filepath: str, remote_chunk_
...
+
+
+def get_downloader(
     remote_dir: str, cache_dir: str, chunks: List[Dict[str, Any]], storage_options: Optional[Dict] = {}
 ) -> Downloader:
     for k, cls in _DOWNLOADERS.items():
         if str(remote_dir).startswith(k):
             return cls(remote_dir, cache_dir, chunks, storage_options)
-    raise ValueError(f"The provided `remote_dir` {remote_dir} doesn't have a downloader associated.")
+    else:
```
tchaton, line 268: Might be better to keep the default behaviour here to avoid surprising behaviour.

src/litdata/streaming/downloader.py
```diff
@@ -239,20 +239,32 @@ def download_file(self, remote_filepath: str, local_filepath: str, remote_chunk_
...
     "azure://": AzureDownloader,
     "hf://": HFDownloader,
     "local:": LocalDownloaderWithCache,
-    "": LocalDownloader,
 }
 
 
-def get_downloader_cls(
+def register_downloader(prefix: str, downloader_cls: type[Downloader], overwrite=False) -> None:
+    if prefix in _DOWNLOADERS and not overwrite:
```
tchaton, line 252: Mind adding some docstring ?

tchaton (commented):
Tiny nits. Thanks ;)

src/litdata/streaming/downloader.py
ouj: Yes. sure.

I thought we were practicing "code as documentation" 🤣

src/litdata/streaming/downloader.py
ouj: As I mentioned in the PR description, this ValueError exception should never be trigger in the old code, because the `"": LocalDownloader,` mapping in the `_DOWNLOADERS`. 

`.startswith("")` will match anything, so any unknown protocol will be treated as `LocalDownloader`. This change just make this explicit.

To make things worse, if we do `_DOWNLOADERS["dummy:"] = DummyDownloader`, it will put `"dummy:"` behind `""` because that is how python dict works. The `for k, cls in _DOWNLOADERS.items():` will stop before it reaches `"dummy:"`, and treat `dummy://something` as a local file folder.


I could keep the exception but removing the `""` mapping, the problem with that is plain folder path like `/home/ouj/dataset` will stop working, and have to be prefixed with `local:/home/ouj/dataset`

tchaton (approved):

Issue #512, opened 2025-03-12
Local cache dir not fully clearing in DDP multi-node training.


JackUrb:
## 🐛 Bug

At the moment it seems that a significant portion of data stored in the cache (~40-60%) ends up not removed over the course of training, and then remains in the cache dir upon completion of training. I suspect this 

### To Reproduce

Currently training a few models over 8 nodes with a large source dataset (only 1 epoch), and the cache dir size accumulates indefinitely. Over the process, it seems that the lock counts for these files are much greater than 0, so I wonder if it has to do with `force_download`-related behavior?

### Expected behavior

The cache dir does not expand significantly above its target size, and files aren't being locked as many times.

### Additional context

<details>
  <summary>Environment detail</summary>

- PyTorch Version (e.g., 1.0):
- OS (e.g., Linux):
- How you installed PyTorch (`conda`, `pip`, source):
- Build command you used (if compiling from source):
- Python version:
- CUDA/cuDNN version:
- GPU models and configuration:
- Any other relevant information:

</details>

tchaton:
Hey @JackUrb. Good idea.

bhimrazy:
Hi @JackUrb, we recently added a few fixes related to cache clearance. It's not released yet, but available on the `main` branch—feel free to give it a try and let us know how it goes. 

Thanks!

JackUrb:
Hi @bhimrazy - launching a new run on litdata `main` this week, will report back.

JackUrb:
@bhimrazy Unfortunately no dice - we end up with an overfull cache directory containing all of the `dat-0-x.bin` from with x between 0 to 40, and then around 1 in 10 of the next 100,000 shards.

The first 40 end up having a `.cnt` of ~22+/-1, and almost all of the rest have 1 with an occasional 2. I suspect the greater in the first ones to be caused by using the same dataset for val as for train. Two separate runs used the same cache, so the 1's and occasional 2 makes sense to me as some race condition occurring roughly 1 in 20 shards.

JackUrb:
Hi @bhimrazy, of course, here's what I've got!
- Number of nodes: 4
- Number of CPU cores and devices per node: 8 GPUs per node, 20 CPUs per GPU
- Number of workers allocated per process: 8-10 workers in a `StreamingDataLoader`
- Cache size allocated per node (in GB): 100GB allocated globally, all pointing to the same cache dir
- Total dataset size (in GB/TB): ~2.2TB
[...]

PR #536, opened 2025-04-01
Fix/last chunk deletion

### LINKED ISSUE #533: Last Chunk is not sent for deletion within each worker group while streaming and it doesnot get deleted at all when in deletion mode.
### Description
The last chunk within each worker group is not sent for deletion while streaming in deletion mode, leading to undeleted chunks. This increases the cache directory size unnecessarily.
> The number of chunks left is direclty proprotional to the number of workers

This issue is also dependent on #532 #534 

#### Affected Code
The issue occurs in the dataset reader, where the final chunk is missing from the deletion process:
[[Reader Code](https://github.com/Lightning-AI/litdata/blob/cfac30acfe3d718540c3b00cc48d24fb1f28917d/src/litdata/streaming/reader.py#L413-L419)](https://github.com/Lightning-AI/litdata/blob/cfac30acfe3d718540c3b00cc48d24fb1f28917d/src/litdata/streaming/reader.py#L413-L419)

This bug is also related to #512 , as it also contributes to increase in cache directory usage.

### Steps to Reproduce
1. Enable deletion mode while streaming a dataset.
2. Observe that the last chunk within each worker group is not deleted.
3. Verify the cache directory size before and after processing.

### Expected Behavior
The last chunk within each worker group should be sent for deletion after processing is completed to prevent unnecessary storage accumulation and should get deleted.

### Code Sample
<details>
  <summary>Minimal Reproducible Example</summary>

```python
# Provide a minimal, self-contained code snippet to reproduce the issue
```

</details>

### Environment
<details>
  <summary>System Details</summary>
  
  - **PyTorch Version**: (e.g., 2.0.1)
  - **OS**: (e.g., Ubuntu 20.04)
  - **Installation Method**: (`pip`, `conda`, source, etc.)
  - **Python Version**: (e.g., 3.10)
  - **CUDA/cuDNN Version**: (if applicable)
  - **GPU Models and Configuration**: (if applicable)
  - **Any other relevant information**:

</details>

bhimrazy:
## What does this PR do?  
Fixes #533.  

This PR fixes the deletion issue for the last chunk within in a worker process, when in `deletion` mode and updates the order in which the `_pre_download_counter` is decremented.  
<details>
<summary> <b>Details on _pre_download_counter </b></summary>

Previously, `_pre_download_counter` was decremented before `_can_delete_chunk()` was evaluated. However, `_can_delete_chunk()` depends on `_pre_download_counter`, leading to incorrect behavior.  

For example, if there is only one chunk and `_pre_download_counter` value is incremented to `1` after download, the counter should allow its deletion later. But because `_pre_download_counter` was already decremented back to `0` before the check, `_can_delete_chunk()` would always return `False`.  

Consider a case where `_max_pre_download = 2`:  
```python
self._pre_download_counter >= self._max_pre_download - 1
```
If `_pre_download_counter` is 0, this evaluates as `0 >= 1`, which is `False`, preventing the chunk from being deleted.  

By adjusting the order of operations, we ensure that `_can_delete_chunk()` gets evaluated with the correct counter value, allowing proper last chunk deletion.  
</details>

Also, part of #512.
> This PR only addresses the deletion of the last remaining chunk. There are still underlying issues with lock count decrements, which seem to appear when the number of ddp processes exceeds 5 (not 100% sure, but observed in a few tests). This prevents the deletion of a lot of chunks.

Experiment results [here](https://github.com/Lightning-AI/litData/pull/536#issuecomment-2782586876)

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/reader.py
+++ b/src/litdata/streaming/reader.py
@@ -170,9 +170,11 @@ class PrepareChunksThread(Thread):
             self._to_delete_queue, timeout=_LONG_DEFAULT_TIMEOUT if reached_pre_download else _DEFAULT_TIMEOUT
         )
 
-        if chunk_index is not None:
-            # Store the current chunk index
-            self._chunks_index_to_be_deleted.append(chunk_index)
+        if chunk_index is None:
+            return
+
+        # Store the current chunk index
+        self._chunks_index_to_be_deleted.append(chunk_index)
 
         # Get the current cache size and decide whether we need to start cleanup. Otherwise, keep track of it
         while self._max_cache_size and self._chunks_index_to_be_deleted and self._can_delete_chunk():
--- a/src/litdata/streaming/reader.py
+++ b/src/litdata/streaming/reader.py
@@ -185,7 +185,8 @@ class PrepareChunksThread(Thread):
 
     def _can_delete_chunk(self) -> bool:
         if self._delete_chunks_when_processed:
-            return self._pre_download_counter >= self._max_pre_download - 1
+            # TODO: check why it is >= self.max_pre_download - 1
+            return self._pre_download_counter >= 0
         return (
             self._max_cache_size is not None
             and _get_folder_size(self._config._cache_dir, self._config) >= self._max_cache_size
--- a/src/litdata/streaming/reader.py
+++ b/src/litdata/streaming/reader.py
@@ -422,6 +422,8 @@ class BinaryReader:
             self._prepare_thread._decrement_local_lock(index.chunk_index)
             self._prepare_thread.delete([index.chunk_index])
             self._prepare_thread.stop()
+            if self._max_cache_size and self._prepare_thread.is_alive():
+                self._prepare_thread.join()
             self._prepare_thread = None
             self._item_loader.close(self._last_chunk_index)
             self._last_chunk_index = None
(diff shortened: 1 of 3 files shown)


bhimrazy:
### Experiments
<details>
  <summary>Screenshots</summary>

image
</details>

| Case | Branch | # of DDP Processes | # of DataLoader Workers per DDP Process | Remaining Cache Size |
|------|--------|---------------------|------------------------------------------|------------------------|
| 1    | main   | 1                   | 32                                       | 1.9 G                |
| 2    | main   | 4                  |8                                       | 1.8 G                |
| 3    | main   | 4                  | 16                                      | 3.7 G                |
| 4    | bhimrazy:fix/last-chunk-deletion   | 1                   | 32                                       | 232 K                |
| 5    |bhimrazy:fix/last-chunk-deletion   | 4                  |8                                       | 232 K                |
| 6    |bhimrazy:fix/last-chunk-deletion   | 4                  | 16                                      | 248 K                |





**Note**: This PR only addresses the deletion of the last remaining chunk. There are still underlying issues with lock count decrements, which seem to appear when the number of ddp processes exceeds 5 (not 100% sure, but observed in a few tests). This prevents the deletion of a lot of chunks and the remaining cache size were around the max_cache size.
<details>
<summary>Experiment with num of ddp process : 6 </summary>

image
</details>

tchaton (approved):
really nice :)

deependujha:
Hi @bhimrazy , great work. But, let's not go for merge right now. It'll be good to compare the two.

PR #574, opened 2025-04-28
Fix dataset generation: deterministic per-index seeding and collate-compatible image format

### LINKED ISSUE #573: Error in the Getting Started Example -- default_collate cannot handle PIL.Images
## 🐛 Bug

Hello! Excited about making use of litdata.optimize to solve some long standing data loading issues Ive encountered many times before.

While trying to get it working with my dataset, I tried running through the getting started example that creates binary files of a fake dataset consisting of random noise images, but pretty quickly ran into an error that should've been caught.

Running `litdata.optimize()` runs without an issue, and the `streaming.py` also technically runs without issue. But, if I try to do anything with the DataLoader defined at the end of `streaming.py`, I get a `default_collate` error because it is incapable of batching PIL.Images.

Given that pretty much all of the examples I could find in the documentation rely heavily on PIL.Image, I'm wondering

A. Can we update the getting started example so that it at least defines a DataLoader that wont immediately throw an error?
B. Can we get some more explanation/exploration into whether it's best to save images as PIL.Image within the optimized files, or if it's better to convert them to torch.Tensor/np.Arrays in advance so they're not repeatedly converted every epoch (My intuition is leaning towards the latter)

### To Reproduce

If desired, I can provide more details on the environment I used to produce this error, but given that it's already based on the minimal working example I anticipate it shouldnt require much imagination to reproduce.

dipeshbabu:
## What does this PR do?

**Issues Addressed**  
1. **Fixes #573**: Two critical bugs in the Getting Started example:  
   - **Duplicate data** when using `optimize` with `num_workers > 1` due to unseeded randomness.  
   - **`default_collate` error** caused by returning PIL Images incompatible with PyTorch batching.  

**Root Causes**  
- **Duplicate Data**: Workers shared the global `numpy` random state, leading to identical random values across processes.  
- **Collate Error**: PIL Images cannot be batched by PyTorch’s `default_collate`.  

**Changes**  
1. **Deterministic Data Generation**:  
   - Seed `numpy`’s RNG uniquely per `index` using `np.random.default_rng(seed=index)`.  
   - Replace `np.random.randint` with the seeded generator’s `rng.integers(...)`.  
2. **Collate Compatibility**:  
   - Return images as **numpy arrays** instead of PIL Images.  
3. **Documentation Updates**:  
   - Updated the Getting Started example to reflect both fixes.  

**Result**  
- No duplicate data across workers.  
- `StreamingDataLoader` now works out-of-the-box with the example.  
- Improved efficiency (no runtime PIL-to-tensor conversions).  

## PR review

## Did you have fun?

### DIFF
--- a/examples/getting_started/generate.py
+++ b/examples/getting_started/generate.py
@@ -10,7 +10,7 @@ def random_images(index):
     return {
         "index": index,  # int data type
         "image": rng.integers(0, 256, (32, 32, 3), dtype=np.uint8),  # numpy array
-        "class": rng.integers(10), # numpy array
+        "class": rng.integers(10),  # numpy array
     }
 
 


bhimrazy:
Hi @dipeshbabu,
The provided sample is just a minimal demo — in real-world use cases, you’d typically work with actual `.jpg` or `.png` images, which get optimized and loaded as tensors directly.

If you’re saving PIL images directly, you can handle them either by subclassing the streaming dataset to apply transforms or by passing a custom `collate_fn` like:

```python
def collate_fn(batch):
    return {
        "image": [sample["image"] for sample in batch],
        "class": [sample["class"] for sample in batch],
    }

train_dataloader = ld.StreamingDataLoader(train_dataset, collate_fn=collate_fn)
```

If you’re interested in making further contributions to **litdata**, we’d be happy to discuss and collaborate on our [Discord](https://discord.com/invite/MWAEvnC5fU) — join us in the **#litdata** channel!
Or please feel free to discuss directly over the issues.

bhimrazy:
Hey @dipeshbabu, just following up — if you're interested, adding the `collate_fn` directly in getting_started/stream.py and any other related places would be a great addition to round out the fix. Would be awesome to have that as part of your first contribution! 🙌

Issue #524, opened 2025-03-24
Cycle option for StreamingDataLoader


Aceticia:
## 🚀 Feature
A function or an argument in StreamingDataLoader to cycle the passed in StreamingDataset.

### Motivation
Many training scenarios in CV involve training models with multiple epochs, while wanting to control the exact number of steps being trained, independent of the underlying dataset size. E.g., given a CombinedStreamingDataset of some length, restart its iterations when it is exhausted.

### Pitch
I'm not quite sure how this should be done - maybe in __iter__ method of StreamingDataLoader, we can catch the final iteration and restart it?

tchaton:
You could check PyTorch Lightning Cycle Loaders: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.utilities.combined_loader.html

philgzl:
As mentioned by @lantiga in #576, the ability to cycle datasets using `ParallelStreamingDataset` is a nice option as it is, but this should probably be upstreamed to `StreamingDataset` in the future.

Idk if this issue should stay open so we don't forget.

philgzl:
Mmh I was thinking of a similar solution to what was implemented in `ParallelStreamingDataset`:

```
dset = ld.StreamingDataset("..", length=100)
```

Iterating over the dataset once then yields 100 samples. If the dataset has less than 100 samples, we cycle and shuffle internally. If we iterate over the dataset a second time, we resume from where we left off without re-shuffling, and yield 100 samples again.

This way we can disentangle the epoch length (as in the number of items yielded by `iter`) from the actual number of samples in the dataset. I believe this is what OP meant.

deependujha:
thanks for the clarification.

Similar to parallelSD, pass int or "inf". Sounds good to me.

philgzl:
Yes and then I guess this feature should be removed from `ParallelStreamingDataset` since we would just pass `StreamingDataset` instances which were already configured to cycle.

PR #576, opened 2025-05-01
Add `ParallelStreamingDataset`

### LINKED ISSUE #554: How to correctly mix multiple `StreamingDataset` to create new data?
I would like to iterate over two `StreamingDataset` instances infinitely and yield new data derived from the output of both datasets. I am able to achieve this by wrapping two `StreamingDataset` instances in an `IterableDataset` as follows:

```python
class MixedDataset(torch.utils.data.IterableDataset):
    def __init__(self, dset_1: ld.StreamingDataset, dset_2: ld.StreamingDataset) -> None:
        self.dset_1 = dset_1
        self.dset_2 = dset_2

    def __iter__(self) -> Iterator[torch.Tensor]:
        dset_iter_1 = iter(self.dset_1)
        dset_iter_2 = iter(self.dset_2)
        while True:
            try:
                x_1 = next(dset_iter_1)
            except StopIteration:
                dset_iter_1 = iter(self.dset_1)
                x_1 = next(dset_iter_1)
            try:
                x_2 = next(dset_iter_2)
            except StopIteration:
                dset_iter_2 = iter(self.dset_2)
                x_2 = next(dset_iter_2)
            yield do_stuff(x_1, x_2)
```

This works but I cannot use this new dataset with a `StreamingDataloader` since it does not inherit `StreamingDataset`. I can use a vanilla `torch.utils.data.DataLoader`, but am I missing out on key features provided by `StreamingDataloader`?

The `CombinedStreamingDataset` does not fit my use case as it yields items from one or the other dataset, from what I understand.

philgzl:
<details>

</details>

## What does this PR do?

Fixes #554.

This PR adds a new dataset class `ParallelStreamingDataset`. As opposed to `CombinedStreamingDataset` which yields a sample from one of the datasets it wraps at each iteration, `ParallelStreamingDataset` yields samples from all the datasets it wraps at each iteration.

This allows to truly combine multiple datasets to generate new data on-the-fly (e.g. multi-source audio signals from different types of sources), which is more flexible compared to pre-generating, optimizing and uploading a frozen version of the combined data.

Since `ParallelStreamingDataset` and `CombinedStreamingDataset` share common functionality, a base class `_BaseStreamingDatasetWrapper` was added.

Instead of adopting a solution similar to `weights` to compensate for the different length of the wrapped datasets, cycling of the wrapped datasets was implemented. Cycling is controlled with the `length` option. If `None`, iteration stops as soon as one of the wrapped datasets is exhausted (no cycling). If an integer or `float("inf")`, the wrapped datasets are cycled until `length` items are yielded. This might solve #524, as we can wrap a `StreamingDataset` with `ParallelStreamingDataset` to cycle it.

## PR review

## Did you have fun?


bhimrazy:
Hi @philgzl ,
Thanks for creating this PR and adding support for parallel streaming! 
Just wanted to share a runtime error I encountered when testing with `batch_size >= 3` and `num_workers >= 3`:

<details><summary>RuntimeError: Failed to get sample from dataset after cycling. Is the dataset empty?</summary>
<p>

```python
from litdata import ParallelStreamingDataset, StreamingDataLoader, StreamingDataset, optimize


def optimize_fn_1(index):
    return index


def optimize_fn_2(index):
    return index + 100


if __name__ == "__main__":
    # optimize(
    #     fn=optimize_fn_1,
    #     inputs=list(range(10)),
    #     output_dir="dataset_1",
    #     num_workers=4,
    #     chunk_bytes="64MB",
    # )

    # optimize(
    #     fn=optimize_fn_2,
    #     inputs=list(range(20)),
    #     output_dir="dataset_2",
    #     num_workers=4,
    #     chunk_bytes="64MB",
    # )

    d1 = StreamingDataset(input_dir="dataset_1")
    d2 = StreamingDataset(input_dir="dataset_2")
    dataset = ParallelStreamingDataset([d1, d2], length=32)
    data_loader = StreamingDataLoader(dataset, batch_size=4, num_workers=4, drop_last=True)

    for batch in data_loader:
        print(batch)
```

```python
  File "/Users/bhimrajyadav/litdata/.venv/lib/python3.12/site-packages/torch/utils/data/_utils/fetch.py", line 33, in fetch
    data.append(next(self.dataset_iter))
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bhimrajyadav/litdata/src/litdata/streaming/parallel.py", line 198, in __next__
    samples, _resets = zip(*[self._get_sample(i) for i in range(len(self._datasets))])
                             ^^^^^^^^^^^^^^^^^^^
  File "/Users/bhimrajyadav/litdata/src/litdata/streaming/parallel.py", line 224, in _get_sample
    raise RuntimeError("Failed to get sample from dataset after cycling. Is the dataset empty?")
RuntimeError: Failed to get sample from dataset after cycling. Is the dataset empty?
```

</p>
</details> 

I’ll keep reviewing and share further feedback if I spot anything else.

philgzl:
I think this is because worker 4 is assigned 0 samples from dataset 1 (dataset length is 10, batch size is 4 and number of workers is 4).

The reasoning behind the RuntimeError was: if dataset raises a StopIteration, then cycle it and try to fetch again. If for some reason it raises a StopIteration again, then raise a RuntimeError. A consequence of this is that the case where some workers are assigned 0 samples is not supported. So maybe we should not raise that RuntimeError and just raise the second StopIteration.

deependujha:
IMO, for exhausted datasets, we need to reset_state_dict() for each dataset worker, so the `__iter__` method doesn't restarts using state_dict.

src/litdata/utilities/base.py
```diff
@@ -0,0 +1,139 @@
...
+        for dataset in self._datasets:
+            dataset.set_drop_last(drop_last)
+
+    def reset_state_dict(self) -> None:
+        """Reset the state of the dataset."""
+        for dataset in self._datasets:
+            dataset.reset_state_dict()
+
+    def _check_datasets(self, datasets: List[StreamingDataset]) -> None:
+        if any(not isinstance(d, StreamingDataset) for d in datasets):
```
deependujha, line 67:
```suggestion
        if any(not isinstance(d, (StreamingDataset, CombinedStreamingDataset)) for d in datasets):
```

I think ParallelSD can also include CombinedSD. Any potential pitfalls with that?

src/litdata/streaming/parallel.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Args:
+            datasets: The list of the StreamingDataset to use.
+            length: The number of samples to yield. If ``None``, the datasets are iterated over until one of them is
+                exhausted. If an integer, the datasets are cycled until ``length`` samples are yielded. Can be
+                ``float("inf")`` for an infinite dataset.
+            force_override_state_dict: Boolean flag for allowing local arguments to override a loaded state dict.
+
+        """
+        self._check_datasets(datasets)
```
deependujha, line 58: I think ParallelSD can also include CombinedSD. Any potential pitfalls with that?

```
def _check_datasets(self, datasets: List[StreamingDataset]) -> None:
        if any(not isinstance(d, (StreamingDataset, CombinedStreamingDataset)) for d in datasets):
            raise RuntimeError("The provided datasets should be instances of the StreamingDataset.")
```

philgzl: That would be cool, I can give it a try

deependujha: - lol, what if, `parallelSD` can take `combinedSD`, &, `combinedSD` can take `parallelSD`!

Screenshot 2025-05-07 at 12 49 07 AM

philgzl:
I think with `lightning.seed_everything` we would still need to compute the right seed from the current dataset state to allow for resuming.

Agree `transform` is probably a better name.

I might need help to fix the tests on macOS :eyes:.

bhimrazy:
For the macOS test, I believe it will require a rerun by the admin to allow it to run beyond 35 minutes, as it's currently exceeding the limit. @philgzl, you can ignore it for now unless an error is reported in the test.

deependujha:
oh sorry, increased it to 45 as tests have increased and to be on the safer side.

deependujha (approved):
Super cool contribution.

Added some minor comments from my side, and once you add `transform` method support, it'll be good to merge.

tests/conftest.py
```diff
@@ -62,6 +62,21 @@ def combined_dataset(tmpdir_factory):
     return CombinedStreamingDataset(datasets=[dataset_1, dataset_2])
 
 
+@pytest.fixture
+def parallel_dataset(tmpdir_factory, request):
```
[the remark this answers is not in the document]
philgzl: 👍 will change

src/litdata/streaming/parallel.py
```diff
@@ -0,0 +1,245 @@
...
+            return
+        for dataset in self._datasets:
+            dataset.set_epoch(current_epoch)
+
+    def get_len(self, num_workers: int, batch_size: int) -> Optional[int]:
+        self.num_workers = num_workers
+        self.batch_size = batch_size
+        lengths = [self._get_len(d) for d in self._datasets]
+        if self._length is None:
+            return min(lengths)
```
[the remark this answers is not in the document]
philgzl: I will triple check but I think it wrote it this way to set the number of workers and the batch size to all the wrapped datasets by calling `_get_len(d)` even if `self._length` is not `None`. If that's the case then I can definitely add a comment.

src/litdata/streaming/parallel.py
```diff
@@ -0,0 +1,245 @@
...
+        except StopIteration as e:
+            if self._length is None:
+                raise e
+            self._dataset_iters[dataset_index] = iter(self._datasets[dataset_index])
+            _reset = True
+            try:
+                sample = next(self._dataset_iters[dataset_index])
+            except StopIteration as e:
+                # The dataset is empty or this worker got 0 samples assigned. Either way raise the StopIteration.
+                raise e
```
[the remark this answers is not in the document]
philgzl: That makes sense. Then the example by @bhimrazy above would not be supported and would raise that RuntimeError. However is that consistent with `StreamingDataset` and `CombinedDataset`? I believe these do support small datasets with too many workers.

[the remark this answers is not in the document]
deependujha: That's an interesting take. @bhimrazy above example dataset has only `10 items` btw. 

Btw, even when we support small datasets, `length` property doesn't work as expected, and stopiteration is raised beforehand.
If I want to infinitely cycle over my small dataset, then it will also fail.

What do you think @bhimrazy ? Which way makes more sense!

src/litdata/streaming/dataloader.py
```diff
@@ -580,10 +583,10 @@ def __init__(
         collate_fn: Optional[Callable] = None,
         **kwargs: Any,
     ) -> None:  # pyright: ignore
-        if not isinstance(dataset, (StreamingDataset, CombinedStreamingDataset)):
+        if not isinstance(dataset, (StreamingDataset, CombinedStreamingDataset, ParallelStreamingDataset)):
```
tchaton, line 586: We could simply check whether it is a StreamingDataset or a _BaseStreamingDatasetWrapper

src/litdata/streaming/dataloader.py
```diff
@@ -631,13 +635,19 @@ def __init__(
...
         if not self.restore:
-            self._latest_worker_idx = 0
-            self._worker_idx = cycle(list(range(self.num_workers if self.num_workers > 0 else 1)))
-            self._worker_idx_iter = iter(self._worker_idx)
+            if (
+                not isinstance(self.dataset, ParallelStreamingDataset)
+                or self.dataset._length is None
+                or self.current_epoch == 0
+            ):
+                self._latest_worker_idx = 0
```
tchaton, line 643: Better to wrap those blocks of code into method, so they can be tested and extended with more ease.

src/litdata/streaming/dataloader.py
```diff
@@ -655,14 +665,30 @@ def __iter__(self) -> Any:
...
+                                if dset.current_epoch != i_cycle + 1:
+                                    dset.current_epoch = i_cycle + 1
+
                     yield batch[__SAMPLES_KEY__]
                 else:
                     yield batch
+
+        # For ParallelStreamingDataset with _length != None we want to cycle the wrapped datasets i.e. we do not want to
+        # restart at index 0 at every epoch. So we set them in restore state.
+        if isinstance(self.dataset, ParallelStreamingDataset) and self.dataset._length is not None:
```
tchaton, line 689: This shouldn't work in theory as the epoch would be the same and it would lead to sampling data in the same order. Let's add a few more tests for fault tolerance to cover this in particular.

src/litdata/streaming/dataloader.py
```diff
@@ -748,14 +793,38 @@ def load_state_dict(self, obj: Dict[str, Any]) -> None:
...
+            self.dataset._set_use_streaming_dataloader(True)
+            self.dataset.load_state_dict(obj)
+            num_samples_yieled, num_cycles = self.dataset._get_num_samples_yielded(
+                self._num_samples_yielded_wrapper, self._num_cycles
+            )
+            if self.dataset._length is None:
+                dset_len = self.dataset._len()
+                if dset_len is not None and any(0 < samples < dset_len for samples in num_samples_yieled):
+                    self.restore = True
+            if self.dataset._length == float("inf") and any(samples > 0 for samples in num_samples_yieled):
```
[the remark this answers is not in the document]
philgzl: I am accessing the `_length` attribute in multiple places to check whether the wrapped datasets are cycling or not.

I can instead add a `.is_cycling()` method that would check `_length` internally. Would that be satisfactory?

src/litdata/streaming/dataloader.py
```diff
@@ -748,14 +793,38 @@ def load_state_dict(self, obj: Dict[str, Any]) -> None:
...
                 self.restore = True
 
+        elif isinstance(self.dataset, ParallelStreamingDataset):
+            self.dataset._set_use_streaming_dataloader(True)
+            self.dataset.load_state_dict(obj)
+            num_samples_yieled, num_cycles = self.dataset._get_num_samples_yielded(
+                self._num_samples_yielded_wrapper, self._num_cycles
+            )
+            if self.dataset._length is None:
+                dset_len = self.dataset._len()
```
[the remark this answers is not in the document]
philgzl: Yes I added this `_len` method as a workaround because `mypy` was complaining about `len()` argument not being sizable since `ParallelStreamingDataset.__len__` can return `None`.

Any idea for an alternative?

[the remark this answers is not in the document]
philgzl: Removed `_len` and now calling `len()` with `# type: ignore` similar to instances of `CombinedStreamingDataset`

src/litdata/streaming/dataloader.py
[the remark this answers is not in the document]
philgzl: Implemented `.is_cycling()` and `.is_infinite()` instead of accessing `_length`

src/litdata/streaming/dataloader.py
philgzl: Done.

Also updated type hint from `Union[StreamingDataset, CombinedStreamingDataset, ParallelStreamingDataset]` to `Union[StreamingDataset, _BaseStreamingDatasetWrapper]`

src/litdata/streaming/dataloader.py
philgzl: We do want the sampling to be the same on the next epoch, since we are resuming in the middle of an internal dataset cycle and the samples in the next epoch should be consistent with the previous epoch (as in from the same cycle). We want to re-shuffle only when starting a new internal dataset cycle.

Unless you meant something else?

Added a test `test_dataloader_shuffle` in `test_parallel.py` which checks we get the behavior described above.

PR #583, opened 2025-05-06
fix: multi-node parquet indexing

### LINKED ISSUE #578: There is still a risk of race condition when indexing an s3 bucket with multi node (Parquet datasets)
There is still a risk of race condition when indexing an s3 bucket with multi node.

Reported: https://github.com/Lightning-AI/litData/pull/569#pullrequestreview-2798291543 by @tchaton

deependujha:
<details>

</details>

## What does this PR do?

Fixes #578 

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/writer.py
+++ b/src/litdata/streaming/writer.py
@@ -31,7 +31,7 @@ from litdata.utilities.encryption import Encryption, EncryptionLevel
 from litdata.utilities.env import _DistributedEnv, _WorkerEnv
 from litdata.utilities.format import _convert_bytes_to_int, _human_readable_bytes
 from litdata.utilities.parquet import get_parquet_indexer_cls
-from litdata.utilities.torch_utils import maybe_barrier
+from litdata.utilities.torch_utils import is_local_rank_0, maybe_barrier
 
 
 @dataclass
@@ -589,10 +589,9 @@ def index_parquet_dataset(
     if not _POLARS_AVAILABLE:
         raise ModuleNotFoundError("Please, run: `pip install polars`")
 
-    env = _DistributedEnv.detect()
-    if env.global_rank == 0:
-        # in distributed mode, only node: 0, process: 0 will create the index
-        # and write it to the cache_dir
+    if is_local_rank_0():
+        # in multi-node setup, each node's first process should do the indexing,
+        # but only first node's first process should upload the index (if uploading is needed)
         pq_chunks_info = []
         config: Dict[str, Any] = {
             "compression": None,
--- a/src/litdata/streaming/writer.py
+++ b/src/litdata/streaming/writer.py
@@ -588,6 +588,11 @@ def index_parquet_dataset(
     if not _POLARS_AVAILABLE:
         raise ModuleNotFoundError("Please, run: `pip install polars`")
 
+    env = _DistributedEnv.detect()
+    if env.global_rank != 0:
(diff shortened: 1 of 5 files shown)


src/litdata/utilities/hf_dataset.py
```diff
@@ -32,28 +30,38 @@ def index_hf_dataset(dataset_url: str, cache_dir: Optional[str] = None) -> str:
...
+    # Otherwise, create a new index file
+    with tempfile.TemporaryDirectory() as temp_dir:
+        temp_index_path = os.path.join(temp_dir, _INDEX_FILENAME)
+        index_parquet_dataset(dataset_url, temp_dir, num_workers=os.cpu_count() or 4)
+
+        # Prepare the cache directory and move the index file there
+        cache_dir = _try_create_cache_dir(dataset_url, cache_dir, index_path=temp_index_path)
+        assert cache_dir is not None
+        if is_local_rank_0():
+            # Only the first process on the first node should create the cache directory
```
deependujha, line 53:
```suggestion
            # Only the first process on the each node should create the cache directory
```

src/litdata/streaming/writer.py
```diff
@@ -588,56 +589,62 @@ def index_parquet_dataset(
...
-            desc="Indexing progress",
-            total=len(pq_dir_class.files),
-            smoothing=0,
-            mininterval=1,
-            leave=True,
-            dynamic_ncols=True,
-            unit="step",
-        )
+    env = _DistributedEnv.detect()
+    if env.global_rank == 0:
```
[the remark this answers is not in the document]
deependujha: nah, only first node should do the indexing, not the rest.

bhimrazy, line 593: @deependujha, could this create a blocker for HF indexing in a multi-node scenario?

bhimrazy (commented):
Nice, @deependujha. It seems like a more solid solution. 

> Maybe, in case it works as a decorator, we could introduce another decorator, something like `local_rank_zero_only`, inspired by [this](https://github.com/Lightning-AI/utilities/blob/main/src/lightning_utilities/core/rank_zero.py).

src/litdata/streaming/writer.py
deependujha: why? Even in multi-node, only one process should do the indexing, then upload index.json.

But, on all the nodes, atleast one process must copy index.json to cache_dir to be used.

Am I missing something?

deependujha:
I also thought of decorator inspired by lightning, but, right now, only interested in blocking a segment of code, and not a function.

Maybe in future, we might introduce it.

src/litdata/streaming/writer.py
bhimrazy: I’m not entirely sure, but it seems so. Perhaps you could run a test for multinode HF indexing in Lightning AI to double check? 
From my understanding, only one node might have the index file, as we don't upload it to the source in the HF scenario, and other processes may be blocked from indexing.

deependujha: great catch, Thanks.

deependujha: tested on multinode in Lightning Studio and it works fine.

deependujha:
For the failing docs in CI

Issue is open in Sphinx repo: https://github.com/sphinx-doc/sphinx/issues/13533

bhimrazy (approved):

PR #597, opened 2025-05-19
Add support for torch.uint16 data type

### LINKED ISSUE #596: Support for uint16 tensors
## 🚀 Feature

My data is in uint16 and I've created uint16 tensors, however this results in a long error, the heart of which is 

```
  File ".venv/lib/python3.11/site-packages/litdata/streaming/serializers.py", line 247, in serialize
    dtype_indice = self._dtype_to_indices[item.dtype]
                   ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: torch.uint16
```

My plan was just to convert to float32 later

### Motivation

uint16 is smaller than float32

### Alternatives

convert to uint8, but I would loose information

bhimrazy:
## What does this PR do ?

Fixes #596
Introduce support for the torch.uint16 data type in the _TORCH_DTYPES_MAPPING.

> test already seems to be present under `test_tensor_serializer`


tchaton (approved):

src/litdata/constants.py
```diff
@@ -74,6 +74,7 @@
     17: torch.int64,
     18: torch.long,
     19: torch.bool,
+    20: torch.uint16,
```
tchaton, line 77: Mind adding a test ?

bhimrazy: @tchaton, it looks like we already have one that covers all the torch data types:
https://github.com/Lightning-AI/litData/blob/32950dd3ec0dd1472c212de8a19c04fe08e7db7f/tests/streaming/test_serializer.py#L204-L223

Do we still need a separate one?

PR #606, opened 2025-05-27
feat: support Queue-based streaming inputs for optimize via new recipe

deependujha:
<details>

</details>

## What does this PR do?

This adds a new`QueueDataChunkRecipe` recipe that enables streaming data into optimize() via a shared queue, rather than requiring all inputs up front.

Useful for reinforcement learning workflows (e.g. DQN, DDPG) or real-time ingestion setups (e.g. from websockets or other async producers).

Built on top of the shared_queue infrastructure. Once merged, we can support memory-efficient, streaming-based optimization flows that weren't previously possible.

```python
from multiprocessing import Process, Queue
from litdata.processing.data_processor import ALL_DONE
import litdata as ld
import time

def yield_numbers():
    for i in range(1000):
        time.sleep(0.01)
        yield (i, i**2)

def data_producer(q: Queue):
    for item in yield_numbers():
        q.put(item)
    q.put(ALL_DONE)  # Sentinel value to signal completion

def fn(index):
    return index  # Identity function for demo

if __name__ == "__main__":
    q = Queue(maxsize=100)

    producer = Process(target=data_producer, args=(q,))
    producer.start()

    ld.optimize(
        fn=fn,                   # Function to process each item
        queue=q,                 # 👈 Stream data from this queue
        output_dir="fast_data",  # Where to store optimized data
        num_workers=2,
        chunk_size=100,
        mode="overwrite",
    )

    producer.join()
```

## PR review

## Did you have fun?

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,11 +127,19 @@ Load the data by replacing the PyTorch DataSet and DataLoader with the Streaming
 ```python
 import litdata as ld
 
-train_dataset = ld.StreamingDataset('s3://my-bucket/fast_data', shuffle=True, drop_last=True)
-train_dataloader = ld.StreamingDataLoader(train_dataset)
+dataset = ld.StreamingDataset('s3://my-bucket/fast_data', shuffle=True, drop_last=True)
 
-for sample in train_dataloader:
-    img, cls = sample['image'], sample['class']
+# Custom collate function to handle the batch (Optional)
+def collate_fn(batch):
+    return {
+        "image": [sample["image"] for sample in batch],
+        "class": [sample["class"] for sample in batch],
+    }
+
+
+dataloader = ld.StreamingDataLoader(dataset, collate_fn=collate_fn)
+for sample in dataloader:
+    img, cls = sample["image"], sample["class"]
 ```
 
 **Key benefits:**
--- a/examples/getting_started/README.md
+++ b/examples/getting_started/README.md
@@ -64,5 +64,16 @@ img = sample['image']
 cls = sample['class']
 
 # Create dataLoader and iterate over it to train your AI models.
-dataloader = StreamingDataLoader(dataset)
+
+# Custom collate function to handle the batch (Optional)
+def collate_fn(batch):
+    return {
+        "image": [sample["image"] for sample in batch],
+        "class": [sample["class"] for sample in batch],
+    }
+
+
+dataloader = StreamingDataLoader(dataset, collate_fn=collate_fn)
+for sample in dataloader:
+    img, cls = sample["image"], sample["class"]
 ```
--- a/examples/getting_started/stream.py
+++ b/examples/getting_started/stream.py
@@ -12,4 +12,16 @@ img = sample["image"]
 cls = sample["class"]
 
 # Create dataLoader and iterate over it to train your AI models.
-dataloader = StreamingDataLoader(dataset)
+
+
+# Custom collate function to handle the batch (Optional)
+def collate_fn(batch):
+    return {
+        "image": [sample["image"] for sample in batch],
+        "class": [sample["class"] for sample in batch],
+    }
+
+
+dataloader = StreamingDataLoader(dataset, collate_fn=collate_fn)
+for sample in dataloader:
+    img, cls = sample["image"], sample["class"]


src/litdata/processing/functions.py
```diff
@@ -419,6 +456,15 @@ def optimize(
...
     _check_version_and_prompt_upgrade(__version__)
 
+    if queue is not None and inputs is not None:
+        raise ValueError("You can either provide `inputs` or `queue`, not both.")
+
+    if queue is None and inputs is None:
+        raise ValueError("At least one of `inputs` or `queue` must be provided.")
+
+    if queue is not None and not isinstance(queue, mp.queues.Queue):
+        raise ValueError(f"The provided `queue` should be a multiprocessing queue. Found {type(queue)}.")
```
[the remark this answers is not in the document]
deependujha: raises this error:
```
File "/private/tmp/lightning/data/main.py", line 25, in <module>
    ld.optimize(
  File "/Users/Deependu/open-source/lightning-ai/litdata/src/litdata/processing/functions.py", line 465, in optimize
    if queue is not None and not isinstance(queue, mp.Queue):
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  ```

So, using the current.

File: src/litdata/processing/data_processor.py [hunk 2]
```diff
@@ -606,10 +609,16 @@ def _loop(self) -> None:
...
 
-            assert isinstance(combined_data, tuple), f"Invalid data received from queue {combined_data=}."
-            assert len(combined_data) == 3, f"Invalid data received from queue {combined_data=}."
+            if self.using_queue_optimize:
+                # using random index number, just to satisfy the default combined_data structure.
+                index = -5  # not using -1, as items will be of length 0, so last_index (0-1 = -1) will become true.
+                item = combined_data
+                paths = None
+            else:
+                assert isinstance(combined_data, tuple), f"Invalid data received from queue {combined_data=}."
```
lantiga, line 618: These `assert`s are user-facing, since they rely on what the user emits from the queue in the processor function, right?
In such case we should favor raising exceptions, and use assert for checking invariants developer-facing.

src/litdata/processing/functions.py
```diff
@@ -351,8 +385,9 @@ def map(
 #
 def optimize(
     fn: Callable[[Any], Any],
-    inputs: Union[Sequence[Any], StreamingDataLoader],
-    output_dir: str,
+    inputs: Optional[Union[Sequence[Any], StreamingDataLoader]] = None,
+    output_dir: str = "optimized_data",
+    queue: Optional[mp.Queue] = None,
```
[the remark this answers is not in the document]
deependujha: Yeah, totally fair point, it could've gone either way.

I kept queue as a separate argument mainly for simplicity and clarity in the function signature. Felt a bit cleaner to separate the streaming mechanism from the static inputs, even if they're functionally similar.

That said, extending inputs to accept a queue is definitely a valid design too, more unified and maybe a bit more intuitive from a caller's perspective.

We can update this in future. But, for now, I think it's good to go.

bhimrazy (approved):
LGTM so far! Let's wait for at least one more approval before merging.

One small note: it would be great to increase test coverage for all the raised `ValueError`s.

src/litdata/processing/data_processor.py
```diff
@@ -606,10 +614,16 @@ def _loop(self) -> None:
                     return
                 continue
 
-            assert isinstance(combined_data, tuple), f"Invalid data received from queue {combined_data=}."
-            assert len(combined_data) == 3, f"Invalid data received from queue {combined_data=}."
+            if self.using_queue_optimize:
+                # using random index number, just to satisfy the default combined_data structure.
+                index = -5  # not using -1, as items will be of length 0, so last_index (0-1 = -1) will become true.
```
tchaton, line 619: This might cause some issues with the cache as it stores the data based on those indexes

tchaton (approved):

src/litdata/processing/data_processor.py
deependujha: IMO, what you are referring to you is: **`Baseworker: self._index_counter = 0`** that tracks number of items added in cache.

It's a separate count maintained by worker and independent of this dummy index.
This dummy index is only used by `_handle_data_transform_recipe` to check if `is_last` index.

tchaton: It would be good to clean up this part. It is a bit hacky

PR #619, opened 2025-06-11
fix: breaking tqdm progress bar in optimizing dataset

### LINKED ISSUE #76: Prints inside the worker processes mess up the progress bar
## 🐛 Bug

In my code, I am enabling a `tqdm` bar per worker with:

```python
    global_rank = int(os.environ["DATA_OPTIMIZER_GLOBAL_RANK"])
    num_workers = int(os.environ["DATA_OPTIMIZER_NUM_WORKERS"])
    local_rank = global_rank % num_workers
    for example in tqdm(data, position=local_rank):
        tokens = tokenizer.encode(example)
        yield tokens
```

But litdata prints this in each rank:

```python
Rank 3 inferred the following `['no_header_tensor:16']` data format.
```

Breaking the `tqdm` bars at the beginning.

Since this print doesn't seem very useful for users, I would suggest that it is removed or put under `fast_dev_run` or a similar verbose-like flag.

deependujha:
<details>

</details>

## What does this PR do?

Fixes #76 

- `before`
Screenshot 2025-06-11 at 8 02 37 PM

- `current fix`
Screenshot 2025-06-12 at 4 11 02 AM

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/processing/data_processor.py
+++ b/src/litdata/processing/data_processor.py
@@ -460,6 +460,7 @@ class BaseWorker:
         worker_index: int,
         num_workers: int,
         node_rank: int,
+        msg_queue: Queue,
         data_recipe: "DataRecipe",
         input_dir: Dir,
         output_dir: Dir,
@@ -485,6 +486,7 @@ class BaseWorker:
         self.worker_index = worker_index
         self.num_workers = num_workers
         self.node_rank = node_rank
+        self.msg_queue = msg_queue
         self.data_recipe = data_recipe
         self.input_dir = input_dir
         self.output_dir = output_dir
@@ -533,7 +535,7 @@ class BaseWorker:
         except Exception:
             traceback_format = traceback.format_exc()
             self.error_queue.put(traceback_format)
-        print(f"Worker {str(_get_node_rank() * self.num_workers + self.worker_index)} is done.")
+        self.msg_queue.put_nowait(f"Worker {str(_get_node_rank() * self.num_workers + self.worker_index)} is done.")
 
     def _setup(self) -> None:
         self._set_environ_variables()
@@ -590,7 +592,9 @@ class BaseWorker:
                     or combined_data == ALL_DONE
                     or (self.keep_data_ordered and num_downloader_finished == self.num_downloaders)
                 ):
-                    print(f"Worker {str(_get_node_rank() * self.num_workers + self.worker_index)} is terminating.")
+                    self.msg_queue.put_nowait(
+                        f"Worker {str(_get_node_rank() * self.num_workers + self.worker_index)} is terminating."
+                    )
 
                     if isinstance(self.data_recipe, DataChunkRecipe):
                         self._handle_data_chunk_recipe_end()
@@ -669,6 +673,7 @@ class BaseWorker:
             encryption=self.data_recipe.encryption,
             writer_chunk_index=self.writer_starting_chunk_index,
             item_loader=self.item_loader,
+            msg_queue=self.msg_queue,
         )
         self.cache._reader._rank = _get_node_rank() * self.num_workers + self.worker_index
 
@@ -1143,6 +1148,9 @@ class DataProcessor:
         self.keep_data_ordered = keep_data_ordered
         self.shared_queue: Union[Queue, FakeQueue, None] = None
 
+        # Queue for routing worker logs to the main process without breaking tqdm output.
+        self.msg_queue: Queue = Queue()
+
         self.state_dict = state_dict or dict.fromkeys(range(self.num_workers), 0)
 
         if self.reader is not None and self.weights is not None:
@@ -1294,6 +1302,19 @@ class DataProcessor:
         total_num_items = len(user_items) if isinstance(user_items, list) else -1
 
         while True:
+            # check if there're msgs in the msg queue
+            msgs = []
+            while not self.msg_queue.empty():
+                msg = self.msg_queue.get()
+                msgs.append(msg)
+
+            if _TQDM_AVAILABLE and len(msgs) > 0:
+                pbar.clear()  # clear the previous progress bar
+            for msg in msgs:
+                print(msg)
+            if _TQDM_AVAILABLE and len(msgs) > 0:
+                pbar.display()  # display the progress bar again
+
             # Exit early if all the workers are done.
             # This means either there were some kinda of errors, or optimize function was very small.
             if all(not w.is_alive() for w in self.workers):
@@ -1336,6 +1357,7 @@ class DataProcessor:
                     json.dump({"progress": str(100 * current_total * num_nodes / total_num_items) + "%"}, f)
 
         if _TQDM_AVAILABLE:
+            pbar.clear()
             pbar.close()
 
         print("Workers are finished.")
@@ -1386,6 +1408,7 @@ class DataProcessor:
                 worker_idx,
                 self.num_workers,
                 _get_node_rank(),
+                self.msg_queue,
                 data_recipe,
                 self.input_dir,
                 self.output_dir,
--- a/src/litdata/streaming/writer.py
+++ b/src/litdata/streaming/writer.py
@@ -16,6 +16,7 @@ import os
 import uuid
 import warnings
 from dataclasses import dataclass
+from multiprocessing import Queue
 from time import sleep, time
 from typing import Any, Dict, List, Optional, Tuple, Union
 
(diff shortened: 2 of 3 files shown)


src/litdata/processing/data_processor.py
```diff
@@ -1294,6 +1302,20 @@ def run(self, data_recipe: DataRecipe) -> None:
         total_num_items = len(user_items) if isinstance(user_items, list) else -1
 
         while True:
+            # check if there're msgs in the msg queue
+            msgs = []
+            while not self.msg_queue.empty():
+                msg = self.msg_queue.get()
```
bhimrazy:
```suggestion
                msg = self.msg_queue.get(timeout=0.001)
```

src/litdata/processing/data_processor.py
```diff
@@ -533,7 +535,7 @@ def run(self) -> None:
         except Exception:
             traceback_format = traceback.format_exc()
             self.error_queue.put(traceback_format)
-        print(f"Worker {str(_get_node_rank() * self.num_workers + self.worker_index)} is done.")
+        self.msg_queue.put_nowait(f"Worker {str(_get_node_rank() * self.num_workers + self.worker_index)} is done.")
```
bhimrazy, line 538: In my case, this part didn’t appear at all — probably it exited early.
<details><summary>Logs</summary>

```python
Create an account on https://lightning.ai/ to optimize your data faster using multiple nodes and large machines.
Setting multiprocessing start_method to spawn. 
Storing the files under /Users/bhimrajyadav/Developer/LightningAI/litdata/fast_data
Setup started with fast_dev_run=False.
Setup finished in 0.003 seconds. Found 10000 items to process.
Starting 8 workers with 10000 items. The progress bar is only updated when a worker finishes.
Workers are ready ! Starting data processing...
Rank 4 inferred the following `['int', 'pil', 'int']` data format.
Rank 0 inferred the following `['int', 'pil', 'int']` data format.                                                                
Rank 3 inferred the following `['int', 'pil', 'int']` data format.                                                                
Rank 5 inferred the following `['int', 'pil', 'int']` data format.                                                                
Rank 1 inferred the following `['int', 'pil', 'int']` data format.                                                                
Rank 2 inferred the following `['int', 'pil', 'int']` data format.
Rank 6 inferred the following `['int', 'pil', 'int']` data format.                                                                
Rank 7 inferred the following `['int', 'pil', 'int']` data format.                                                                
Worker 0 is terminating.                                                                                                          
Worker 3 is terminating.                                                                                                          
Worker 4 is terminating.                                                                                                          
Worker 6 is terminating.                                                                                                          
Worker 1 is terminating.                                                                                                          
Worker 5 is terminating.                                                                                                          
Worker 7 is terminating.                                                                                                          
Worker 2 is terminating.                                                                                                          
Progress: 100%|████████████████████████████████████████████████████████████████████████████| 10000/10000 [00:19<00:00, 514.59it/s]
Workers are finished.                                                                                                             
Finished data processing!
```
</details> 

<details><summary>Code</summary>

```python
import numpy as np
from PIL import Image
import litdata as ld


def random_images(index):
    # Replace with your actual image loading here (e.g., .jpg, .png, etc.)
    # (recommended to pass as compressed formats like JPEG for better storage and optimized streaming speed)
    # You can also apply resizing or reduce image quality to further increase streaming speed and save space.
    fake_images = Image.fromarray(np.random.randint(0, 256, (128, 128, 3), dtype=np.uint8))
    fake_labels = np.random.randint(10)

    # You can use any key:value pairs. Note that their types must not change between samples, and Python lists must
    # always contain the same number of elements with the same types.
    data = {"index": index, "image": fake_images, "class": fake_labels}

    return data


if __name__ == "__main__":
    # The optimize function writes data in an optimized format.
    ld.optimize(
        fn=random_images,  # the function applied to each input
        inputs=list(range(10000)),  # the inputs to the function (here it's a list of numbers)
        output_dir="fast_data",  # optimized data is stored here
        num_workers=8,  # The number of workers on the same machine
        chunk_bytes="64MB",  # size of each chunk
    )
```
</details>

src/litdata/streaming/writer.py
```diff
@@ -89,6 +92,8 @@ def __init__(
         self._compression = compression
         self._encryption = encryption
         self._item_loader = item_loader or PyTreeLoader()
+        if msg_queue is not None:
+            self.msg_queue = msg_queue
```
bhimrazy:
```suggestion
        self.msg_queue = msg_queue
```

src/litdata/streaming/writer.py
```diff
@@ -173,7 +178,11 @@ def serialize(self, items: Any) -> Tuple[bytes, Optional[int]]:
 
             worker_rank = get_worker_rank()
             if worker_rank is not None:
-                print(f"Rank {worker_rank} inferred the following `{data_format}` data format.")
+                msg = f"Rank {worker_rank} inferred the following `{data_format}` data format."
+                if hasattr(self, "msg_queue"):
```
bhimrazy:
```suggestion
                if self.msg_queue is not None:
```

src/litdata/processing/data_processor.py
deependujha: nice catch. It was completed before all the msgs could be printed. Ensured waiting.

Please check again and verify.

src/litdata/processing/data_processor.py
```diff
@@ -1335,7 +1360,26 @@ def run(self, data_recipe: DataRecipe) -> None:
...
+                break
+
+        if len(msgs) > 0:
+            if _TQDM_AVAILABLE:
+                pbar.clear()  # clear the previous progress bar
+            for msg in msgs:
+                print(msg)
+            if _TQDM_AVAILABLE:
+                pbar.display()  # display the progress bar again
+
```
bhimrazy: This message-draining and printing logic seems to appear multiple times — 
it might be a good to move it into a helper method like self._drain_msgs() 
(or another suitable name) to improve clarity and reduce repetition.

src/litdata/processing/data_processor.py
```diff
@@ -1335,7 +1345,10 @@ def run(self, data_recipe: DataRecipe) -> None:
                 with open("status.json", "w") as f:
                     json.dump({"progress": str(100 * current_total * num_nodes / total_num_items) + "%"}, f)
 
+        flush_msg_queue(self.msg_queue, pbar if _TQDM_AVAILABLE else None)
+
         if _TQDM_AVAILABLE:
+            pbar.clear()
             pbar.close()
 
         print("Workers are finished.")
```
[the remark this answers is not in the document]
deependujha: it leaves another progress bar

src/litdata/processing/data_processor.py
```diff
@@ -1602,3 +1616,31 @@ def in_notebook() -> bool:
     shell.
     """
     return "ipykernel" in sys.modules
+
+
+def flush_msg_queue(msg_queue: Queue, pbar: Optional[Any] = None):
+    """Flush messages from a queue and print them without breaking the tqdm progress bar.
```
bhimrazy, line 1622:
```suggestion

def _flush_msg_queue(msg_queue: Queue, pbar: Optional[Any] = None):
    """Flush messages from a queue and print them without breaking the tqdm progress bar.
```

let's keep it as internal

src/litdata/processing/data_processor.py
```diff
@@ -1602,3 +1616,31 @@ def in_notebook() -> bool:
...
+            msgs.append(msg)
+        except Empty:
+            break
+    if len(msgs) > 0:
+        if _TQDM_AVAILABLE:
+            pbar.clear()  # clear the previous progress bar
+        for msg in msgs:
+            print(msg)
+        if _TQDM_AVAILABLE:
+            pbar.display()  # display the progress bar again
```
bhimrazy: I think here we should check for `pbar` and whether the required attributes are available, instead of relying on `_TQDM_AVAILABLE`,
for example: `if pbar and hasattr(pbar, "clear"): pbar.clear()`.

bhimrazy (commented):
Almost there — just added a few final comments.

src/litdata/processing/data_processor.py
deependujha: This is just a helper function, and not a private method. I don't find any reason to mark it as internal/private.

If we go by this logic, then all the utility function has to be marked with `_`.

src/litdata/processing/data_processor.py
deependujha:
```python
        if pbar is not None:
            pbar.clear()  # clear the previous progress bar
        for msg in msgs:
            print(msg)
        if pbar is not None:
            pbar.display()  # display the progress bar again
```

I think, rather I should've used this. If you're using this helper function, then it assumes you've passed tqdm progress bar. 

**Allowing anything with a clear method to pass would make bugs harder to detect if an invalid object is passed in**.

PR #631, opened 2025-06-17
feat: fast random access for streamingDataset without chunk downloading

### LINKED ISSUE #14: Fast random access for `StreamingDataset`
## 🚀 Feature

Support a way to request just a single sample from a StreamingDataset without internally pulling the whole chunk.

### Motivation

Streaming chunks is great for cases where you want to visit the whole dataset but sub-optimal if you just want to view individual samples. Right now, if you just index a StreamingDataset directly the latency is very high. This is a bit of an issue if you want to explore the dataset (e.g. in a streamlit or gradio app).

### Pitch

We could have a way to request a single sample from the dataset that would download only the bytes of that sample instead of downloading the whole chunk. This would enable building visualizations etc. on top of streaming datasets.

deependujha:
<details>

</details>

## What does this PR do?

Fixes #14 

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/cache.py
+++ b/src/litdata/streaming/cache.py
@@ -50,6 +50,7 @@ class Cache:
         session_options: Optional[Dict] = {},
         max_pre_download: int = 2,
         msg_queue: Optional[Queue] = None,
+        no_chunk_download: bool = False,
     ):
         """The Cache enables to optimise dataset format for cloud training. This is done by grouping several elements
         together in order to accelerate fetching.
@@ -70,6 +71,7 @@ class Cache:
             session_options: Additional options for the S3 session.
             max_pre_download: Maximum number of chunks that can be pre-downloaded while filling up the cache.
             msg_queue: Optional message queue to send messages to the main process.
+            no_chunk_download: If True, fetch only the requested sample's bytes instead of downloading the entire chunk.
 
         """
         super().__init__()
@@ -100,6 +102,7 @@ class Cache:
             storage_options=storage_options,
             session_options=session_options,
             max_pre_download=max_pre_download,
+            no_chunk_download=no_chunk_download,
         )
         self._is_done = False
         self._distributed_env = _DistributedEnv.detect()
--- a/src/litdata/streaming/config.py
+++ b/src/litdata/streaming/config.py
@@ -151,6 +151,26 @@ class ChunksConfig:
 
         self.try_decompress(local_chunkpath)
 
+    def download_chunk_bytes_from_index(self, chunk_index: int, offset: int, length: int) -> bytes:
+        assert self._chunks is not None
+        chunk_filename = self._chunks[chunk_index]["filename"]
+
+        local_chunkpath = os.path.join(self._cache_dir, chunk_filename)
+
+        if os.path.exists(local_chunkpath):
+            raise ValueError("`download_chunk_bytes_from_index` should not be called for already downloaded chunks.")
+
+        if self._compressor is not None:
+            raise ValueError(
+                "The `download_chunk_bytes_from_index` method is not supported for compressed chunks. "
+                "Please, use `download_chunk_from_index` instead."
+            )
+
+        if self._downloader is None:
+            raise RuntimeError("The downloader is not initialized. Please, initialize it before downloading chunks.")
+
+        return self._downloader.download_chunk_bytes_from_index(chunk_index, offset, length)
+
     def try_decompress(self, local_chunkpath: str) -> None:
         if self._compressor is None:
             return
--- a/src/litdata/streaming/dataset.py
+++ b/src/litdata/streaming/dataset.py
@@ -63,6 +63,7 @@ class StreamingDataset(IterableDataset):
         index_path: Optional[str] = None,
         force_override_state_dict: bool = False,
         transform: Optional[Callable] = None,
+        no_chunk_download: bool = False,
     ) -> None:
         """The streaming dataset can be used once your data have been optimised using the DatasetOptimiser class.
 
@@ -90,6 +91,7 @@ class StreamingDataset(IterableDataset):
                 If `index_path` is a full file path, it will use that directly.
             force_override_state_dict: Boolean flag for allowing local arguments to override a loaded state dict.
             transform: Optional transformation function to apply to each item in the dataset.
+            no_chunk_download: If True, fetch only the requested sample's bytes instead of downloading the entire chunk.
         """
         _check_version_and_prompt_upgrade(__version__)
 
@@ -201,6 +203,7 @@ class StreamingDataset(IterableDataset):
             if not callable(transform):
                 raise ValueError(f"Transform should be a callable. Found {transform}")
             self.transform = transform
+        self.no_chunk_download = no_chunk_download
 
     def set_shuffle(self, shuffle: bool) -> None:
         self.shuffle = shuffle
@@ -240,6 +243,7 @@ class StreamingDataset(IterableDataset):
             storage_options=self.storage_options,
             session_options=self.session_options,
             max_pre_download=self.max_pre_download,
+            no_chunk_download=self.no_chunk_download,
         )
         cache._reader._try_load_config()
 
--- a/src/litdata/streaming/downloader.py
+++ b/src/litdata/streaming/downloader.py
@@ -76,8 +76,20 @@ class Downloader(ABC):
 
         logger.debug(_get_log_msg({"name": f"download_chunk_from_index_{chunk_index}", "ph": "E"}))
 
+    def download_chunk_bytes_from_index(self, chunk_index: int, offset: int, length: int) -> None:
+        chunk_filename = self._chunks[chunk_index]["filename"]
+        remote_chunkpath = os.path.join(self._remote_dir, chunk_filename)
+
+        return self.download_bytes(remote_chunkpath, offset, length)
+
     def download_file(self, remote_chunkpath: str, local_chunkpath: str) -> None:
         pass
+
+    def download_bytes(self, remote_chunkpath: str, offset: int, length: int) -> bytes:
+        raise NotImplementedError(
+            f"`download_bytes` method is not implemented for the {self.__class__.__name__} class. "
+            "Please use a specific downloader implementation that supports byte range downloads."
+        )
 
 
 class S3Downloader(Downloader):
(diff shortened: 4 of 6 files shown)


src/litdata/streaming/cache.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         session_options: Optional[Dict] = {},
         max_pre_download: int = 2,
         msg_queue: Optional[Queue] = None,
+        no_chunk_download: bool = False,
```
tchaton: no_store = False

It does download but it doesn't store the bytes

src/litdata/streaming/reader.py
```diff
@@ -397,9 +401,19 @@ def read(self, index: ChunkedIndex) -> Any:
...
-            )
+            if self.no_chunk_download and self._config and self._config._remote_dir:
+                # we need to download relevant bytes only
+                # then deserialize them, and reconstruct the item with pytree's unflatten function
+                offset = (1 + (index.index - begin) if index.index >= begin else index.index + 1) * 4
+                pair = self.config.download_chunk_bytes_from_index(index.chunk_index, offset, 8)
+                begin, end = np.frombuffer(pair, np.uint32)
+                actual_item_length = end - begin
+                raw_bytes = self.config.download_chunk_bytes_from_index(index.chunk_index, begin, actual_item_length)
+                item = self._item_loader.load_item_from_bytes(raw_bytes, index.chunk_index)
```
bhimrazy: I think it would be nice to keep this entire part inside the item loader. wdyt ?

deependujha: I also wanted the same, but downloading requires `config`, and `config` itself imports `item_loader`, so it'll lead to circular import issue.

I'm thinking of having this logic in a separate helper function in `reader` only.

src/litdata/streaming/cache.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         session_options: Optional[Dict] = {},
         max_pre_download: int = 2,
         msg_queue: Optional[Queue] = None,
+        no_store: bool = False,
```
Borda: This is not a very descriptive argument, not sure what would be better...

From the brainstormed list, here are the top three argument names for the parameter that fetches only the requested sample's bytes instead of the entire chunk:
- `partial_sample_fetch` — Descriptive and reads well in boolean logic...
- `sample_bytes` — Compact, intuitive, and aligns with naming patterns like chunk_bytes.
- `on_demand_bytes` — Communicates purpose clearly and feels modern—great for signaling optimized behavior.

deependujha: Thanks. Went with `on_demand_bytes`, feels the clearest and leaves room for similar options later.

src/litdata/streaming/config.py
```diff
@@ -151,6 +151,28 @@ def download_chunk_from_index(self, chunk_index: int, skip_lock: bool = False) -
...
+        if os.path.exists(local_chunkpath):
+            with open(local_chunkpath, "rb") as f:
+                f.seek(offset)
+                return f.read(length)
+
+        if self._compressor is not None:
+            raise ValueError(
+                "The `download_chunk_bytes_from_index` method is not supported for compressed chunks. "
+                "Please, use `download_chunk_from_index` instead."
+            )
```
bhimrazy, line 169: I wonder if this error message is helpful enough for the user.

src/litdata/streaming/dataset.py
```diff
@@ -444,6 +461,7 @@ def __next__(self) -> Any:
             if self.num_chunks is not None and self.worker_next_chunk_index >= self.num_chunks:
                 self.current_epoch += 1
                 self.reset_state_dict()
+                self.on_demand_bytes = True  # reset on_demand_bytes to True
```
bhimrazy, line 464: Do we need this reset here inside `__next__` ?

src/litdata/streaming/dataset.py
```diff
@@ -287,6 +301,7 @@ def __iter__(self) -> "StreamingDataset":
         self.worker_env = _WorkerEnv.detect()
         self.cache = self._create_cache(worker_env=self.worker_env)
         self.shuffler = self._create_shuffler(self.cache)
+        self.on_demand_bytes = False  # reset on_demand_bytes to False, and store chunks in the cache
```
bhimrazy, line 304: It looks like the cache is being created here first when in iterator mode—does that mean cache is always initialized with `on_demand_bytes=True` ? Just trying to understand the flow.

src/litdata/streaming/downloader.py
```diff
@@ -76,9 +76,21 @@ def download_chunk_from_index(self, chunk_index: int) -> None:
...
+
     def download_file(self, remote_chunkpath: str, local_chunkpath: str) -> None:
         pass
 
+    def download_bytes(self, remote_chunkpath: str, offset: int, length: int) -> bytes:
+        raise NotImplementedError(
+            f"`download_bytes` method is not implemented for the {self.__class__.__name__} class. "
+            "Please use a specific downloader implementation that supports byte range downloads."
+        )
+
```
bhimrazy: I don’t see a flag that allows the user to opt out `on_demand_bytes`. In that case, could this fail for other downloaders that don’t support it? Just want to confirm the intended behavior.

bhimrazy (commented):
Looking good so far!

I’ve added a few comments and questions.
It would also be helpful to include benchmarks comparing a few individual item loads with and without `on_demand_bytes` in the description.

src/litdata/streaming/config.py
deependujha: this error is not intended for the user itself. Rather when someone is implementing a new `item_loader`, then he might encounter this error and may decide to take the relevant action.

src/litdata/streaming/dataset.py
deependujha: I think yes, if you streamed your data and then again want to do random access for visualization. When you're doing multiple iterations or epochs, it will be again set to false in `__iter__`.

src/litdata/streaming/downloader.py
deependujha:
```python
        if isinstance(self._item_loader, PyTreeLoader):
            if (
                self.on_demand_bytes
                and self._config
                and self._config._remote_dir
                and self._config._config
                and not self._config._config.get("encryption", None)
                and not self._config._config.get("compression", None)
            ):
                raw_bytes = self.read_item_bytes(index, begin)
                item = self._item_loader.load_item_from_bytes(raw_bytes, index.chunk_index)
            else:
                self.setup_thread_and_download_chunk(index)
                item = self._item_loader.load_item_from_chunk(
                    index.index, index.chunk_index, chunk_filepath, begin, filesize_bytes, self._encryption
                )
        else:
            self.setup_thread_and_download_chunk(index)
            item = self._item_loader.load_item_from_chunk(
                index.index, index.chunk_index, chunk_filepath, begin, filesize_bytes
            )
```

only for datasets that use `pytreeloader` with no compression and no encryption, use this.

src/litdata/streaming/dataset.py
deependujha: sorry, couldn't understand you. Can you rephrase it!

Btw, a rough summary:

- we have a flag `on_demand_bytes` when turned on, `reader` will download relevant bytes only and not the complete chunk.
- So, when `indexing`, `on_demand_bytes` is set to True (default), and when `__iter__ (iterating)`, we set it to False, so reader downloads the completes chunk and writes in file.
- reader does byte downloading only for datasets with `item_loader: pytreeloader` and no encryption and no compression.

src/litdata/streaming/downloader.py
bhimrazy: I mean, will it still work if we perform index reads directly from Azure for chunks? I didn’t see `download_bytes` implemented— will it fallback?

src/litdata/streaming/dataset.py
bhimrazy: Makes sense. 

After the iterations, maybe we might not need to download the bytes again if they’re already in the cache—unless the cache gets cleared during the process.

deependujha: yeah, but it depends on the cache_size. If the chunk exists, it'll be simply read from there (in byte & chunk download both).

src/litdata/streaming/downloader.py
deependujha: nice catch, lol.

bhimrazy: There’s one more case to consider: since we support registering custom downloaders, users implementing their own downloaders might face the same issue as well.

src/litdata/streaming/dataset.py
bhimrazy: Oh okay, I missed this here. 
So setting `on_demand_bytes = False` also updates the cache’s reader state accordingly.

https://github.com/Lightning-AI/litData/pull/631/files#diff-0bf44049f047c93e7c836b8cabdab89d688b6655a66a8c185527d8f529e69215R205-R216

```python
    @on_demand_bytes.setter
    def on_demand_bytes(self, value: bool) -> None:
        if not isinstance(value, bool):
            raise ValueError(f"on_demand_bytes should be a boolean. Found {value}")
        self._on_demand_bytes = value
        assert self.cache is not None, "Cache must be initialized before setting on_demand_bytes."
        self.cache._reader.on_demand_bytes = value
```

This one is serving two purpose here.  not sure if this is a correct way to do so ?

src/litdata/streaming/downloader.py
deependujha: I've implemented fallback. Shouldn't be an issue now. Please verify.

src/litdata/streaming/dataset.py
deependujha: what would be better approach in your opinion? My thinking was: `dataset is higher level abstraction than reader, so just setting this up should manage the internals`.

src/litdata/streaming/downloader.py
bhimrazy: Looks good.

src/litdata/streaming/dataset.py
bhimrazy: Let’s see if @borda has any suggestions—otherwise, I think we’re good to go.

deependujha: merging this for now. If He recommends something, we can implement it in next pr.

PR #635, opened 2025-06-24
feat(streaming): enable per-dataset batch-sizes in `CombinedStreamingDataset`

### LINKED ISSUE #327: Use different batch sizes in CombinedStreamingDataset
## 🚀 Feature

`CombinedStreamingDataset` allows you to combine multiple `StreamingDataset`s with a sampling ratio -- but it assumes that that the `batch_size` is the same for each dataset. 

### Motivation

If the different datasets have tensors of different sizes, it would be great to use different batch sizes per dataset to maximize throughput / memory consumption (e.g. batch size of 1 for dataset with larger input tensors, batch size of 2 for dataset with smaller input tensors).

### Pitch

Allow `set_batch_size` to take a list of batch_sizes -- one per dataset.

### Alternatives

One thing that that would need to be considered would be gradient accumulation. For example, if dataset A is large tensors, with only 1 fitting in memory per batch and dataset B has small tensors, with 4 fitting in memory per batch, you would want to do 4 steps of gradient accumulation when acting on samples from dataset A if you want a 50-50 split during training between dataset A and dataset B. If you want a different ratio samples from dataset A vs. dataset B, you'd need to be able to make this number of gradient accumulation steps configurable.

MagellaX:
Closes #327.
What does this PR do?

New capability – different batch-sizes per stream
CombinedStreamingDataset.set_batch_size(...) now accepts either
a single int (old behaviour), or
a Sequence[int] with one entry per wrapped StreamingDataset.
_CombinedDatasetIterator (for batching_method="per_stream") keeps yielding from the
selected stream until its own batch-size quota is reached, then switches to a new stream.
Implementation highlights
Core logic lives in:
src/litdata/utilities/base.py – validation & propagation of a Sequence[int].
src/litdata/streaming/combined.py – dataset-specific limit calculation and switching.
Single-int paths are untouched → 100 % backward-compatible.

Tests
Added test_combined_dataset_per_dataset_batch_size in tests/streaming/test_combined.py.
Parametrised with multiple batch-size lists; asserts that no stream emits more samples than its allotment.
All existing tests still pass (pytest -q).

Why is this useful?
Projects combining datasets with very different tensor sizes can now maximise GPU utilisation by
using a smaller batch on the “large” dataset and a larger one on the “small” dataset, without
resorting to gradient-accumulation work-arounds.

### DIFF
--- a/src/litdata/streaming/combined.py
+++ b/src/litdata/streaming/combined.py
@@ -15,11 +15,7 @@ import logging
 import random
 from collections.abc import Iterator, Sequence
 from copy import deepcopy
-
-from typing import Any, Iterator, Literal, Optional, Sequence, Union
-
-from typing import Any, Literal, Optional
-
+from typing import Any, Literal, Optional, Union
 
 from litdata.debugger import ChromeTraceColors, _get_log_msg
 from litdata.streaming.dataset import StreamingDataset
@@ -191,11 +187,10 @@ class _CombinedDatasetIterator(Iterator):
         # a specific batch size per dataset.
         self._batch_size = batch_size
         from collections.abc import Sequence as _Sequence
+
         # Validate when a sequence is provided
         if isinstance(batch_size, _Sequence) and len(batch_size) != len(datasets):
-            raise ValueError(
-                "When providing a sequence of batch sizes, its length must match the number of datasets."
-            )
+            raise ValueError("When providing a sequence of batch sizes, its length must match the number of datasets.")
         self._is_done = False
 
         if num_samples_yielded is not None:
@@ -266,11 +261,7 @@ class _CombinedDatasetIterator(Iterator):
             dataset_idx = self._cur_dataset_index
 
             # Determine the batch-size limit for the current dataset
-            limit = (
-                self._batch_size[dataset_idx]
-                if isinstance(self._batch_size, _Sequence)
-                else self._batch_size
-            )
+            limit = self._batch_size[dataset_idx] if isinstance(self._batch_size, _Sequence) else self._batch_size
 
             if self._samples_yielded_in_batch >= limit:
                 # Current dataset reached its quota; pick a *different* dataset if possible
--- a/src/litdata/utilities/base.py
+++ b/src/litdata/utilities/base.py
@@ -12,12 +12,8 @@
 # limitations under the License.
 
 from abc import ABC, abstractmethod
-
-from typing import Any, Dict, Iterator, List, Optional, Sequence, Union
-
-from collections.abc import Iterator
-from typing import Any, Optional
-
+from collections.abc import Iterator, Sequence
+from typing import Any, Optional, Union
 
 from torch.utils.data import IterableDataset
 
@@ -45,7 +41,7 @@ class _BaseStreamingDatasetWrapper(IterableDataset, ABC):
         for dataset in self._datasets:
             dataset.set_shuffle(shuffle)
 
-    def set_batch_size(self, batch_size: Union[int, Sequence[int]]) -> None:  # noqa: D401
+    def set_batch_size(self, batch_size: Union[int, Sequence[int]]) -> None:
         """Set the current batch size.
 
         This method now supports either:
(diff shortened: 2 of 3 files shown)


deependujha:
Hi @MagellaX 

Thanks for contributing to **LitData**! 🙌
Could you please take a look at the failing checks and fix the issues? Let me know if you need any help.

tchaton:
Hey @MagellaX. Would you mind to fix the failing tests ?

bhimrazy:
Hi @MagellaX,
I've merged the latest changes from `main` to bring this branch up to date.
Whenever you get a chance, please pull the changes into your local branch before making any new commits to avoid conflicts.

Also, I'm reviewing the PR now 😊.

src/litdata/streaming/combined.py
```diff
@@ -239,11 +247,35 @@ def _get_dataset_index(self) -> int:
...
+
+            # Determine the batch-size limit for the current dataset
+            limit = self._batch_size[dataset_idx] if isinstance(self._batch_size, _Sequence) else self._batch_size
+
+            if self._samples_yielded_in_batch >= limit:
+                # Current dataset reached its quota; pick a *different* dataset if possible
+                candidate_idx = self._cur_dataset_index
+                if len([i for i in self._dataset_indexes if i is not None]) > 1:
+                    while candidate_idx == self._cur_dataset_index:
+                        candidate_idx = self._set_new_dataset_index()
```
bhimrazy, line 271: Is there a specific reason we're performing this check inside the loop?
Wouldn't `self._set_new_dataset_index()` already ensure a different index is returned?

src/litdata/streaming/combined.py
```diff
@@ -239,11 +247,35 @@ def _get_dataset_index(self) -> int:
...
+                # Update tracking
+                self._cur_dataset_index = candidate_idx
+                self._samples_yielded_in_batch = 0
+                dataset_idx = self._cur_dataset_index
+                # Re-compute limit for the new dataset
+                if isinstance(self._batch_size, _Sequence):
+                    limit = self._batch_size[dataset_idx]
```
bhimrazy, line 278: Probably we can remove this — it doesn't seem to be used.

src/litdata/utilities/base.py
```diff
@@ -40,11 +40,31 @@ def set_shuffle(self, shuffle: bool) -> None:
         for dataset in self._datasets:
             dataset.set_shuffle(shuffle)
 
-    def set_batch_size(self, batch_size: int) -> None:
-        """Set the current batch size to the datasets."""
-        self.batch_size = batch_size
-        for dataset in self._datasets:
-            dataset.set_batch_size(batch_size)
+    def set_batch_size(self, batch_size: Union[int, Sequence[int]]) -> None:
```
bhimrazy, line 44: One thing to take care of is that the batch size is set from the DataLoader, which will always override any batch size set elsewhere.

@tchaton , any thoughts over this.

src/litdata/streaming/combined.py
MagellaX: _set_new_dataset_index() draws once from the weight distribution—it can legally return the same dataset we were just using.
The inner while … only runs when we have >1 remaining streams; it makes sure we actually switch streams once the current one has hit its per-stream quota. Otherwise the run-length test in test_combined_dataset_per_dataset_batch_size can still be violated.
So the explicit check is intentional, not redundant.

src/litdata/streaming/combined.py
MagellaX: Good catch. After the last refactor, that local limit isn’t referenced anymore. I’ll drop or, I guess, you can do it(edits are allowed by maintainers), those two lines in a follow-up commit to keep things tidy.

src/litdata/utilities/base.py
MagellaX: Correct-the DataLoader still sets a scalar int, so existing behaviour is unchanged.
The wrapper’s new Sequence[int] support is purely for advanced users who call set_batch_size([...]) themselves (e.g., when constructing a CombinedStreamingDataset outside of the Lightning DataLoader). In the DataLoader path, we still receive an int, which is propagated unchanged.
I’ve also added type-safe casting where we forward the value to keep 'mypy' happy.😊

bhimrazy: > I’ve also added type-safe casting where we forward the value to keep 'mypy' happy.😊

btw, I believe there have been some updates to the `mypy` version in the main branch. It would be great if you could update this branch accordingly.
I tried doing it earlier, but the changes were overwritten due to later force pushes.

tchaton (approved):

PR #650, opened 2025-07-04
Add resume option to `ParallelStreamingDataset`

### LINKED ISSUE #649: `ParallelStreamingDataset` does not resume when breaking early
## 🐛 Bug

Currently `ParallelStreamingDataset` can be set to resume from where it left off in the previous epoch by setting `length` to an `int` or `float("inf")`. However this does not work when breaking out of the loop early. If the loop is exited early, the same samples are yielded in the next `iter()`.

This is because the state loading and saving is done internally at the end of `iter()`, instead of at the beginning.

This results in unexpected results when e.g. using `limit_<stage>_batches` in Lightning Trainer.

### To Reproduce

<details>

```python
from litdata import ParallelStreamingDataset, StreamingDataLoader, StreamingDataset
from litdata.streaming import Cache

cache = Cache(input_dir="temp/", chunk_size=1)
dset_len = 10
for i in range(dset_len):
    cache[i] = i
cache.done()
cache.merge()

dset = ParallelStreamingDataset([StreamingDataset("temp/")], length=999)
assert dset.is_cycling()

dloader = StreamingDataLoader(dset)

batches = []
for i,(batch,) in enumerate(dloader):
    batches.append(batch)
    if i == 3:
        break

for i, (batch,) in enumerate(dloader):
    assert batch != batches[i]  # fails
    if i == 3:
        break
```

</details>

### Note

Setting `length` to an `int` or `float("inf")` currently does two things: cycling the datasets to yield `length` samples in `iter()`, and toggling resuming from the previous epoch.

There are cases where the user might want to cycle to yield `length` samples, but not resume in the next epoch. For example for validation or testing datasets, for which the samples should usually be the same across different `iter()` calls.

IMO these two aspects should be separated. This can be done by adding an extra `resume` option to `ParallelStreamingDataset`.

philgzl:
<details>

</details>

## What does this PR do?

Fixes #649.

- Automatic state loading and saving is now done at the beginning of `iter()` instead of at the end.
- Automatic state loading and saving is now controlled by a new option `resume`.
- Refactored `ParallelStreamingDataset` tests a bit.

## PR review

## Did you have fun?

### DIFF
--- a/tests/streaming/test_parallel.py
+++ b/tests/streaming/test_parallel.py
@@ -870,3 +870,42 @@ def test_parallel_dataset_partial_iteration_resume(tmp_path_factory, length, res
             assert all(torch.equal(x, y) for x, y in zip(batch, batches_1[i]))
         if i == break_at:
             break
+
+
+@pytest.mark.parametrize("length", [None, 6])
+@pytest.mark.parametrize("resume", [False, True])
+@pytest.mark.parametrize("shuffle", [False, True])
+def test_parallel_dataset_complete_iteration_resume(tmp_path_factory, length, resume, shuffle):
+    _, _, pardset, dloader = prepare_parallel_dataset_and_dataloder(
+        tmp_path_factory, parlen=length, len1=4, len2=4, batch_size=1, num_workers=2, shuffle=shuffle, resume=resume
+    )
+    assert pardset.is_cycling() or length is None
+    expected_1 = [
+        [torch.tensor([0]), torch.tensor([0])],
+        [torch.tensor([2]), torch.tensor([2])],
+        [torch.tensor([1]), torch.tensor([1])],
+        [torch.tensor([3]), torch.tensor([3])],
+        [torch.tensor([0]), torch.tensor([0])],
+        [torch.tensor([2]), torch.tensor([2])],
+    ]
+    batches_1 = []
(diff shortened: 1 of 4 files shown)


tests/streaming/test_parallel.py
```diff
@@ -542,38 +562,146 @@ def test_parallel_dataset_dataloader_states_partial_iterations(parallel_dataset,
...
+    epoch_1_expected_dataset0_samples_yielded,
+    epoch_1_expected_dataset1_samples_yielded,
+    epoch_2_expected_num_samples_yielded,
+    epoch_2_expected_num_cycles,
+    epoch_2_expected_current_epoch,
+    epoch_2_expected_dataset_1_current_epoch,
+    epoch_2_expected_dataset_2_current_epoch,
+    epoch_2_expected_latest_worker_idx,
+    epoch_2_expected_dataset0_samples_yielded,
+    epoch_2_expected_dataset1_samples_yielded,
```
Borda:
```suggestion
    epoch_1_expected,
    epoch_2_expected,
```
Could we rather have it as two dataClass so it i=easier to follow that each epoch has the same expected fields

philgzl: Done :+1:

Borda (approved):

PR #652, opened 2025-07-06
feat(litdata): Add Support for `StreamingRawDataset` to Stream Raw Datasets from Cloud Storage

bhimrazy:
### Overview
This PR introduces a new `StreamingRawDataset` class designed for efficient streaming of raw files directly from cloud storage (e.g., S3, GCS). 

Unlike optimized formats (e.g., litdata chunks), this class allows users to stream raw datasets (e.g., images, text files, etc) without requiring prior preprocessing or data conversion. It supports scalable, on-the-fly access to data, making it well-suited for training or inference workflows where immediate access to raw files is needed.

> *Note: This is a beta feature and may be subject to minor changes in future updates based on feedback and pending follow-ups.*

### Usage Example

```python
from litdata.streaming.raw_dataset import StreamingRawDataset
from torch.utils.data import DataLoader

# Initialize the streaming raw dataset from S3 path
dataset = StreamingRawDataset("s3://bucket/files/") # also accepts transform here

# Access a single file (raw bytes)
first_sample = dataset[0]
print(f"Type of first sample: {type(first_sample)}")  # return bytes

# Access multiple files by indices using __getitems__
indices = [0, 1, 2]
samples = dataset.__getitems__(indices) # returns a list of 'bytes'
print(f"Retrieved {len(samples)} samples. Type of first sample: {type(samples[0])}")

# Use PyTorch DataLoader to batch iterate over the dataset
dataloader = DataLoader(dataset, batch_size=32)

for batch_idx, batch in enumerate(dataloader):
    print(f"Batch {batch_idx} size: {len(batch)}")
    # Each item in batch is raw bytes of a file
    # You can process batch here...

    # For demo, only process first batch
    break
    
 
# output
# Type of first sample: <class 'bytes'>
# Retrieved 3 samples. Type of first sample: <class 'bytes'>
# Batch 0 size: 32
```

---

### ImageNet Benchmarks

```python
# without transform
python benchmarks/stream_raw_imagenet.py --bytes --epoch 1 

# with transform
python benchmarks/stream_raw_imagenet.py --epoch 1 
```

image

---
**Follow-ups**
- Add support for grouping files
- Add support for storing index file to the source cloud storage
- Switch indexing from `fsspec` to `obstore`
- Move the file indexer to its own folder.
- Check for caching files
- Test for azure and other scenarios

## PR review

## Did you have fun?


src/litdata/streaming/raw_dataset.py
```diff
@@ -0,0 +1,382 @@
...
+            return file_path  # Return remote path as fallback
+
+
+class StreamingRawDataset(IterableDataset):
+    """Stream raw files from cloud storage with fast indexing and caching.
+
+    Supports ImageFolder-style datasets:
+
+    s3://bucket/dataset/
+    ├── class_1/
```
tchaton: We shouldn't overfit to images and let the users get the raw bytes and group the files as he wishes

bhimrazy: Sure @tchaton — thanks for the note.
I didn’t quite get the “grouping” part. Mind elaborating a bit? 
> folder structure, or something different?

src/litdata/streaming/raw_dataset.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+@dataclass(slots=True)
+class FileMetadata:
+    """Metadata for a single file in the dataset."""
+
+    path: str
+    size: int
+    modified_time: Optional[float] = None
+    metadata: dict[str, Any] = field(default_factory=dict)
```
bhimrazy: these can be removed

src/litdata/streaming/downloader.py
```diff
@@ -203,6 +209,56 @@ def download_bytes(self, remote_filepath: str, offset: int, length: int, local_c
...
+
+        if obj.scheme != "s3":
+            raise ValueError(f"Expected obj.scheme to be `s3`, instead, got {obj.scheme} for remote={remote_filepath}")
+
+        bucket = obj.netloc
+        key = obj.path.lstrip("/")
+
+        store = self._get_store(bucket)
+        resp = await obs.get_async(store, key)
+        stream = resp.stream(min_chunk_size=2 * 1024 * 1024)
```
tchaton: I think we should use https://developmentseed.org/obstore/latest/api/get/#obstore.GetResult.buffer_async instead. So we let rust get the bytes ready and then there is a no copy back to python

src/litdata/streaming/downloader.py
```diff
@@ -203,6 +209,56 @@ def download_bytes(self, remote_filepath: str, offset: int, length: int, local_c
...
+                raise ModuleNotFoundError(str(_OBSTORE_AVAILABLE))
+            from obstore.auth.boto3 import Boto3CredentialProvider
+            from obstore.store import S3Store
+
+            session = boto3.Session(**self._storage_options, **self.session_options)
+            credential_provider = Boto3CredentialProvider(session)
+            self._store = S3Store(bucket, credential_provider=credential_provider)
+        return self._store
+
+    async def adownload_fileobj(self, remote_filepath: str, fileobj: Any) -> None:
```
tchaton: I think it is better to change the API here to take the remote_filepath and return the bytes directly.

src/litdata/streaming/downloader.py
tchaton: Screenshot 2025-07-25 at 10 38 09

bhimrazy: Sure, done. One thing we might need to take into account is that it now returns `obstore.Bytes`.

benchmarks/litdata/stream_raw_imagenet.py
```diff
@@ -0,0 +1,144 @@
+"""Benchmark streaming raw dataset with litdata."""
+
+import argparse
```
tchaton: Let's add this under a benchmark folder.

src/litdata/streaming/downloader.py
```diff
@@ -171,18 +181,14 @@ def download_file(self, remote_filepath: str, local_filepath: str) -> None:
                         )
                     raise RuntimeError(error_message)
             else:
```
tchaton: Why did we change this function ?

src/litdata/streaming/raw_dataset.py
```diff
@@ -0,0 +1,365 @@
...
+            with open(os.path.join(index_path), "wb") as f:
+                f.write(zstd.compress(json.dumps(metadata).encode("utf-8")))
+        except (OSError, zstd.ZstdError) as e:
+            logger.warning(f"Error caching index to {index_path}: {e}")
+
+        logger.info(f"Built index with {len(files)} files from {input_dir} at {index_path}")
+        return files
+
+
+class FileIndexer(BaseIndexer):
```
tchaton, line 110: Let's move the file indexer to its own folder.

benchmarks/litdata/stream_raw_imagenet.py
```diff
@@ -0,0 +1,144 @@
+"""Benchmark streaming raw dataset with litdata."""
```
tchaton: Would you mind adding a section on the README for this dataset. 

Basically, when would you want to this vs StreamingDataset.

src/litdata/constants.py
```diff
@@ -44,6 +44,8 @@
 _PIL_AVAILABLE = RequirementCache("PIL")
 _TORCH_VISION_AVAILABLE = RequirementCache("torchvision")
 _AV_AVAILABLE = RequirementCache("av")
+_OBSTORE_AVAILABLE = RequirementCache("obstore")
```
tchaton, line 47: Let's add to the requirements.

benchmarks/litdata/stream_raw_imagenet.py
bhimrazy: Sure

src/litdata/streaming/downloader.py
bhimrazy: you mean `download_fileobj`. It's currently used for getitem. I will switch it to use `async`.

benchmarks/litdata/stream_raw_imagenet.py
bhimrazy: sure.

src/litdata/constants.py
bhimrazy: sure.

benchmarks/litdata/stream_raw_imagenet.py
```diff
@@ -0,0 +1,144 @@
...
+    return T.Compose(
+        [
+            T.Lambda(deserialize_jpeg),
+            T.Resize((224, 224)),
+            T.ToDtype(dtype, scale=True),
+        ]
+    )
+
+
+def main():
```
Borda: Why not use the Lightning CLI?

tchaton:
> Looks great @bhimrazy 💪
> 
> My only ~bikeshed~ observation is around the name: `Raw` makes sense if you know about litdata chunks, but for first time users `Raw` is all they know.
> 
> I'm wondering if a better API would be to not give the user the problem of choosing between `StreamingRawDataset` and `StreamingDataset`, but have a single `StreamingDataset` class handle both "raw" and optimized. After all we know if we're pointing to an optimized dataset or a non-optimized one.
> 
> So you can start using a non-optimized one right away, then you can decide to optimize it and your code will get faster without switching classes. Does this make sense?

Yes, we will likely consolidate this in the future. But for the time being, I would like to land this one to get early feedbacks.

src/litdata/streaming/downloader.py
tchaton: Sorry, I meant change this function.

benchmarks/litdata/stream_raw_imagenet.py
```diff
@@ -0,0 +1,144 @@
...
+    """Deserialize JPEG bytes to a tensor."""
+    arr = torch.frombuffer(data, dtype=torch.uint8)
+    with suppress(RuntimeError):
+        return decode_jpeg(arr, mode=ImageReadMode.RGB)  # RGB
+
+    import io
+
+    from PIL import Image
+
+    img = Image.open(io.BytesIO(data)).convert("RGB")
```
tchaton: Torchvision supports image_read and does this logic under the hood I believe.

benchmarks/litdata/stream_raw_imagenet.py
bhimrazy: Thank you for the suggestion, @Borda! 

I think the `Lightning CLI` is best suited for end-to-end workflows involving the model, datamodule, and training loop.

For pure dataloader speed benchmarking, it feels a bit heavyweight since it requires a complete LightningModule and Trainer setup.

If there’s a simpler way to use Lightning CLI just for dataloader speed, I’d be very interested to learn about it!

src/litdata/streaming/raw_dataset.py
bhimrazy: Sure, keeping this for the follow-up and have added it to the list above in the description.

benchmarks/litdata/stream_raw_imagenet.py
bhimrazy: added `decode_image`.  [`read_image`](https://docs.pytorch.org/vision/main/generated/torchvision.io.read_image.html) seems to take the path.

benchmarks/litdata/stream_raw_imagenet.py
Borda: but then we would need to maintain multiple entry points :(

src/litdata/streaming/downloader.py
```diff
@@ -22,6 +22,7 @@
 from typing import Any, Optional
 from urllib import parse
 
+import boto3
```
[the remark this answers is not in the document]
bhimrazy: can be ignored

src/litdata/streaming/downloader.py
```diff
@@ -171,18 +181,14 @@ def download_file(self, remote_filepath: str, local_filepath: str) -> None:
...
-
                 extra_args: dict[str, Any] = {}
 
                 if not os.path.exists(local_filepath):
-                    # Issue: https://github.com/boto/boto3/issues/3113
                     self._client.client.download_file(
                         obj.netloc,
                         obj.path.lstrip("/"),
                         local_filepath,
                         ExtraArgs=extra_args,
```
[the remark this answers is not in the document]
bhimrazy: Reverted the change — my bad, I missed removing it. It's not in scope.

src/litdata/streaming/raw_dataset.py
```diff
@@ -0,0 +1,365 @@
...
+
+        # Cache the index with ZSTD compression
+        # TODO: upload the index to cloud storage
+        try:
+            metadata = {
+                "source": input_dir,
+                "files": [file.to_dict() for file in files],
+                "created_at": time.time(),
+            }
+            with open(os.path.join(index_path), "wb") as f:
```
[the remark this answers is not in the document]
bhimrazy: Thank you for flagging.

src/litdata/streaming/downloader.py
bhimrazy: Ah, got it — my bad. Thanks for flagging it. I’ve reverted it; I had missed removing it.

src/litdata/streaming/downloader.py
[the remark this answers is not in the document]
bhimrazy: updated

tchaton (approved):
Looks really neat !

benchmarks/litdata/stream_raw_imagenet.py
bhimrazy: Sure, we can definitely look into the possibilities! Maybe we can touch on this briefly in this week's OSS call — would be great to get a bit of insight from you on it.

cc: @deependujha

src/litdata/streaming/raw_dataset.py
```diff
@@ -0,0 +1,365 @@
...
+
+    def _run_async(self, coro: Any) -> Any:
+        """Runs a coroutine, attaching to an existing event loop if one is running."""
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+
+        return loop.run_until_complete(coro)
```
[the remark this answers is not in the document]
bhimrazy: Looking into it, will get back in the follow-ups.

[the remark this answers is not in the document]
bhimrazy: Even the suggested change doesn't seem to be safe. 
Analysis from grok and gemini 2.5 pro were similar and more:

- Python’s asyncio is designed for single-threaded concurrency. Event loops are not thread-safe, and calling asyncio.set_event_loop() from multiple threads can corrupt the asyncio state or raise exceptions.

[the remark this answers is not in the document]
bhimrazy: Below are the main issues with the `_run_async` method in the `StreamingRawDataset` class:

- **Improper Event Loop Management**: Using deprecated `asyncio.get_event_loop()` and creating new loops risks conflicts in environments with running loops, like Jupyter or FastAPI.
- **Lack of Thread Safety**: Calling `asyncio.set_event_loop()` from multiple threads can cause race conditions or corrupt the asyncio state in multi-threaded data loaders.
- **Resource Leakage**: Failing to close newly created event loops after use can lead to resource leaks, causing warnings or memory issues in long-running applications.
- **Poor Integration with Asynchronous Applications**: Managing its own loop disrupts async applications, causing errors when `run_until_complete()` is called in a running loop context.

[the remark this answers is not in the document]
bhimrazy: Recommended Soln after analysis from Gemini 2.5 pro and grok: Replace `_run_async` with a direct call to `asyncio.run()` in `__getitems__`.

Pros:
- Simplifies the code by removing manual event loop management.
- asyncio.run() creates a new event loop, runs the coroutine, and closes the loop automatically, addressing the resource leakage issue.
- Ideal for synchronous callers (e.g., standard PyTorch training loops).

PR #651, opened 2025-07-05
[WIP]: Integrate Ultralytics Support

deependujha:
<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)
- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/lit-data/blob/main/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes # (issue).

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in GitHub issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #655, opened 2025-07-14
Feat: add support to pass transform kwargs in StreamingDataset

deependujha:
<details>

</details>

## What does this PR do?

This PR introduces support for passing keyword arguments (`transform_kwargs`) to transform functions defined in the `StreamingDataset`. This allows users to inject shared configuration, context, or state into their transformation logic without relying on closures, subclassing, or global variables.

### 🧩 Why is this useful?

In many real-world pipelines, transform functions need access to external information like:

* Dataset-specific configurations (`img_size`, `mode`, `names`)
* Parsing methods from dataset classes (e.g., Ultralytics-style label parsers)
* Shared tools or processing logic (e.g., tokenizers, label maps)

This PR enables users to supply those via a clean, declarative API using the `transform_kwargs` argument.

---

### ✅ Key Features

* `transform_kwargs` is a new optional argument to `StreamingDataset`
* Automatically injected into every transform function (or list of functions) via `**kwargs`
* Includes `index` by default for per-sample awareness
* Supports both standalone functions and method-based transforms

---

### 💡 Real-world use case

This feature was motivated by integrating Ultralytics datasets into a streaming pipeline. These datasets contain methods and attributes needed for parsing and postprocessing, such as:

* `.names`, `.is_coco_format`, `.imgsz`
* `custom_label_parser`, `postprocess`

With `transform_kwargs`, we can directly pass these into transform functions, enabling modular and reusable streaming transforms without needing to subclass `StreamingDataset`.

---

### 🧪 Example

```python
def transform_fn(x, **kwargs):
    names = kwargs["names"]
    index = kwargs["index"]
    return parse_and_transform(index, names, x)

dataset = StreamingDataset(
    data_dir,
    cache_dir=str(cache_dir),
    transform=transform_fn,
    transform_kwargs={"names": yolo_dataset.names}
)
```

> `index` is automatically injected even if not specified in `transform_kwargs`.

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/dataset.py
+++ b/src/litdata/streaming/dataset.py
@@ -10,7 +10,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import logging
 import os
 from time import time
@@ -29,7 +28,12 @@ from litdata.streaming.resolver import Dir, _resolve_dir
 from litdata.streaming.sampler import ChunkedIndex
 from litdata.streaming.serializers import Serializer
 from litdata.streaming.shuffle import FullShuffle, NoShuffle, Shuffle
-from litdata.utilities.dataset_utilities import _should_replace_path, _try_create_cache_dir, subsample_streaming_dataset
+from litdata.utilities.dataset_utilities import (
+    _should_replace_path,
+    _try_create_cache_dir,
+    fn_accepts_kwargs,
+    subsample_streaming_dataset,
+)
 from litdata.utilities.encryption import Encryption
 from litdata.utilities.env import _DistributedEnv, _is_in_dataloader_worker, _WorkerEnv
 from litdata.utilities.format import _convert_bytes_to_int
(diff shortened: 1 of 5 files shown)


src/litdata/streaming/dataset.py
```diff
@@ -62,7 +66,10 @@ def __init__(
         max_pre_download: int = 2,
         index_path: Optional[str] = None,
         force_override_state_dict: bool = False,
-        transform: Optional[Callable] = None,
+        transform: Optional[Union[Callable, list[Callable]]] = None,
+        transform_kwargs: Optional[dict[str, Any]] = None,
```
Borda: Do we really need these kwargs? Why not just function and if yser need some custom staff he would pass `transform=partial(my_transform, **transform_kwargs)`

deependujha: haha, my bad. Didn't gave it much thought and went ahead with the first thing that came to mind.
`partial` is much simpler and better. Thanks again for the great catch!

---

I'll revise this PR to only add `support for chaining multiple transform_fn`. That part still feels valuable and clean.

Borda (approved):

deependujha:
Screenshot 2025-07-14 at 9 57 47 PM

https://github.com/Lightning-AI/litData/pull/651/files#diff-7d756c9b6b80c4ccc46eb78e1f37bade5d944e57095e98ea6dcd3b9884d1d51aR76-R80

Ultralytics parses labels file and has some of the helper methods implemented for it. I could've written another wrapper transform method for the same. But, this felt cleaner at that time and I guess might be helpful in future integration as well.

tchaton (approved):

PR #660, opened 2025-07-27
chore: Improve JPEG deserialization to handle PNG images with JPEG extensions

bhimrazy:
# What does this PR do ?

Updates JPEG deserialization to gracefully handle PNG images mislabeled with JPEG extensions, using `torchvision` instead of`PIL`.

### DIFF
--- a/src/litdata/streaming/serializers.py
+++ b/src/litdata/streaming/serializers.py
@@ -129,21 +129,16 @@ class JPEGSerializer(Serializer):
         raise TypeError(f"The provided item should be of type `JpegImageFile`. Found {item}.")
 
     def deserialize(self, data: bytes) -> torch.Tensor:
-        from torchvision.io import decode_jpeg
-        from torchvision.transforms.functional import pil_to_tensor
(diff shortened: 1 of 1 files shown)


src/litdata/streaming/serializers.py
```diff
@@ -129,21 +129,16 @@ def serialize(self, item: Any) -> tuple[bytes, Optional[str]]:
...
 
-        # Fallback to PIL
-        if not _PIL_AVAILABLE:
-            raise ModuleNotFoundError("PIL is required. Run `pip install pillow`")
-        from PIL import Image
-
-        img = Image.open(io.BytesIO(data))
-        return pil_to_tensor(img)
+        # Fallback: decode as a generic image (handles PNG, etc.)
+        return decode_image(array, mode=ImageReadMode.RGB)
```
bhimrazy: To check if we should add `mode` here or not — but I think it would be nice to have.

bhimrazy: Okay, this might not be ideal for cases where the user explicitly needs the original set of channels, as it forces to 3. So, reverting the change.

PR #665, opened 2025-07-31
feat(litdata): Add grouping functionality in StreamingRawDataset

### LINKED ISSUE #662: Add support for re-grouping in Streaming raw dataset
**Notes from @tchaton** 

We would need something like this:

```
class CustomStreamingRawDataset(StreamingRawDataset):

    def setup(self, filepaths: List[str]) -> List[List[str]]:
        ....

        # Note: Each group is a single sample with contains multiple path
        return grouped_filepaths
```

bhimrazy:
## What does this PR do ?

Enhance the StreamingRawDataset class with grouping capabilities in the setup method, allowing for flexible item structuring for users. Improve asynchronous handling for batch downloads and update tests to ensure functionality. Remove unused imports and streamline the code.

Fixes #662 
> Follow up to #652

### Usage Example

```python
from litdata.streaming.raw_dataset import StreamingRawDataset, FileMetadata
from torch.utils.data import DataLoader
from typing import Union

class CustomStreamingRawDataset(StreamingRawDataset):
    def setup(self, files: list[FileMetadata]) -> Union[list[FileMetadata], list[list[FileMetadata]]]:
       # Customize grouping logic here.
       # For example: return pairs like [[image_1, mask_1], [image_2, mask_2], ...]
        return files

# Initialize the streaming raw dataset from S3 path
dataset = CustomStreamingRawDataset("s3://bucket/files/")
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ class SegmentationRawDataset(StreamingRawDataset):
 
 # Initialize the custom dataset
 dataset = SegmentationRawDataset("s3://bucket/files/")
-loader = DataLoader(dataset, batch_size=4)
+loader = DataLoader(dataset, batch_size=32)
 for item in loader:
     # Each item in the batch is a pair: [image_bytes, mask_bytes]
     pass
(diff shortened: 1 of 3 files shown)


tchaton (commented):
Mind adding example in the README ?

bhimrazy:
> Mind adding example in the README ?
Sure! Just updated. 
My bad — I missed adding it earlier. Thanks for the reminder!

tchaton (approved):
Looks great !

PR #671, opened 2025-08-04
ref: Move raw dataset code to `litdata/raw`, expose `StreamingRawDataset` at top-level

bhimrazy:
## What does this PR do ?

This PR reorganizes all raw dataset-related files into the new `litdata/raw` directory and moves corresponding tests to `tests/raw`.

Also, exposes StreamingRawDataset in the top-level `litdata` import for easier access.

> No functional changes; only refactoring and import improvements.

### DIFF
--- a/src/litdata/raw/indexer.py
+++ b/src/litdata/raw/indexer.py
@@ -108,10 +108,10 @@ class FileIndexer(BaseIndexer):
         """Discover dataset files and return their metadata."""
         parsed_url = urlparse(input_dir)
 
-        if parsed_url.scheme in _SUPPORTED_PROVIDERS:
+        if parsed_url.scheme in _SUPPORTED_PROVIDERS:  # Cloud storage
             return self._discover_cloud_files(input_dir, storage_options)
 
-        if not parsed_url.scheme or parsed_url.scheme == "file":
+        if not parsed_url.scheme:  # Local filesystem
             return self._discover_local_files(input_dir)
 
         raise ValueError(
--- a/src/litdata/raw/indexer.py
+++ b/src/litdata/raw/indexer.py
@@ -1,0 +1,178 @@
+# Copyright The Lightning AI team.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


src/litdata/raw/indexer.py
```diff
@@ -0,0 +1,178 @@
...
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Optional
+from urllib.parse import urlparse
+
+from litdata.constants import _FSSPEC_AVAILABLE, _TQDM_AVAILABLE, _ZSTD_AVAILABLE
+
+logger = logging.getLogger(__name__)
+_SUPPORTED_PROVIDERS = ("s3", "gs", "azure")
```
deependujha, line 26: We should move this to `constant.py` file

src/litdata/raw/__init__.py
deependujha, line 1: While not strictly required, it’s generally a good practice to explicitly export relevant classes and helper functions in the package’s `__init__.py` via `__all__`.

```python
__all__ = [
    "BaseIndexer",
    "StreamingRawDataset",
    # etc.
]
```

src/litdata/raw/indexer.py
bhimrazy: yeah, I noticed that `constant.py` already have one being used for other purposes, so I thought it might be better to colocate this one here instead.

deependujha: Even in that case, imo it would be better to have something like `_RAW_DATASET_SUPPORTED_PROVIDERS` in constant.py file.

But, anyways, I guess `constant.py: _SUPPORTED_PROVIDERS` can itself be updated & used. Azure is already supported.

src/litdata/raw/__init__.py
bhimrazy: Thank you, @deependujha — I like the approach. I’ll go ahead and add this for the stable and exposable ones, and we can bring in the rest as they become more stable over time.

deependujha (approved):

src/litdata/raw/indexer.py
bhimrazy: Thank you for the suggestion @deependujha. 

> It looks like we don’t yet have an implementation for Azure — https://github.com/Lightning-AI/litData/blob/main/src/litdata/streaming/fs_provider.py.

That said, defining something like `_RAW_DATASET_SUPPORTED_PROVIDERS` sounds good if we plan to use it across multiple files. Since it's currently only used in this particular file, perhaps we can keep it colocated for now and move it to a shared constants once it’s needed more broadly.

PR #683, opened 2025-08-08
streaming(reader): cleanup chunk lock files by prefix during delete

### LINKED ISSUE #682: Streaming cache leaves stale compressed lock files (e.g., *.zstd.bin.lock ) after chunk deletion
## 🐛 Bug

When streaming compressed datasets (e.g., zstd) from remote storage (S3), the downloader creates a lock file for the compressed chunk (e.g., `chunk-0-56459.zstd.bin.lock`). After download, the chunk is decompressed to `chunk-0-56459.bin`. When cache cleanup later deletes the decompressed `.bin` file, the corresponding compressed `.lock` file is not removed. Over time this leaves many empty `*.bin.lock` and `*.<codec>.bin.lock` files in the cache, which is particularly problematic on network filesystems.

Root cause:
- `filelock` releases the lock on context exit but does not remove the `.lock` file.
- Decompression deletes the compressed file but not its lock.
- Reader’s cleanup only deleted `.bin.lock` and `.cnt.lock` for the uncompressed base, missing the compressed lock filename (e.g., `.zstd.bin.lock`).

### To Reproduce

1. Prepare a dataset with compressed chunk filenames in `index.json`, e.g. `chunk-0-*.zstd.bin`.
2. Use `BinaryReader`/dataloader to stream from S3 (or any remote) with compression enabled (e.g., `zstd`).
3. Set a small cache size to force cleanup during iteration so the decompressed `.bin` files get deleted.
4. Iterate enough to download and delete several chunks.
5. Inspect the cache directory; observe `*.zstd.bin.lock` files remaining even after the corresponding `.bin` was removed.

### Expected behavior

- All lock files associated with a chunk—both the uncompressed base (`chunk-*.bin.lock`, `chunk-*.cnt.lock`) and the compressed variant (`chunk-*.<codec>.bin.lock`)—are removed when the chunk is deleted from cache.
- No stale `*.lock` files remain after chunks are cleaned up.

jwills:
Fixes #682 

Centralizes chunk lock-file cleanup in Reader’s delete path and makes it compression-agnostic by removing any .lock files matching the chunk’s base filename prefix. Adds a test that simulates a non-local downloader which leaves lockfiles in place to ensure cleanup is effective.

Did you have fun?
I think gpt5 enjoyed itself, and if gpt5 is happy, I'm happy

### DIFF
--- a/src/litdata/streaming/reader.py
+++ b/src/litdata/streaming/reader.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import glob
 import logging
 import os
 import warnings
@@ -150,13 +151,13 @@ class PrepareChunksThread(Thread):
         if _DEBUG:
             print(f"Deleted {chunk_filepath} by {self._rank or 0}. Debug: {can_delete_chunk}")
 
-        for lock_extension in [".lock", ".cnt.lock"]:
-            try:
-                locak_chunk_path = chunk_filepath + lock_extension
-                if os.path.exists(locak_chunk_path):
-                    os.remove(locak_chunk_path)
-            except FileNotFoundError:
-                pass
+        base_name = os.path.basename(chunk_filepath)
+        base_prefix = os.path.splitext(base_name)[0]
+        cache_dir = os.path.dirname(chunk_filepath)
+        pattern = os.path.join(cache_dir, f"{base_prefix}*.lock")
+        for lock_path in glob.glob(pattern):
+            with suppress(FileNotFoundError, PermissionError):
+                os.remove(lock_path)
 
     def stop(self) -> None:
         """Receive the list of the chunk indices to download for the current epoch."""
--- a/tests/streaming/test_lock_cleanup.py
+++ b/tests/streaming/test_lock_cleanup.py
@@ -1,0 +1,81 @@
+import os
+import shutil
+from contextlib import suppress
+
+import pytest
+from filelock import FileLock, Timeout
+
+from litdata.constants import _ZSTD_AVAILABLE
+from litdata.streaming.cache import Cache
+from litdata.streaming.config import ChunkedIndex
+from litdata.streaming.downloader import LocalDownloader, register_downloader, unregister_downloader
+from litdata.streaming.reader import BinaryReader
+from litdata.streaming.resolver import Dir
+
+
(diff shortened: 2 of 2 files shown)


src/litdata/streaming/reader.py
```diff
@@ -150,13 +151,13 @@ def _apply_delete(self, chunk_index: int) -> None:
...
-                if os.path.exists(locak_chunk_path):
-                    os.remove(locak_chunk_path)
-            except FileNotFoundError:
-                pass
+        base_name = os.path.basename(chunk_filepath)
+        base_prefix = os.path.splitext(base_name)[0]
+        cache_dir = os.path.dirname(chunk_filepath)
+        pattern = os.path.join(cache_dir, f"{base_prefix}*.lock")
+        for lock_path in glob.glob(pattern):
+            with suppress(FileNotFoundError, PermissionError):
```
deependujha, line 159: why suppressing `permissionError`?

jwills: ack I told gpt5 not to do that but it snuck one past me-- will fix

jwills: ah the model claims it was following the pattern in the file at e.g. https://github.com/Lightning-AI/litData/blob/5f8ea890b70fe6ce46b4c8294c776c3250737ab5/src/litdata/streaming/reader.py#L124 <-- do you want to do it the same way we for those lock files, or should I just check for the `FileNotFoundError`?

deependujha: i don't know why we need to suppress `PermissionError`. My approach will be to even remove `PermissionError` from `reader.py`.

Is there any specific reason we need to suppress `PermissionError` @bhimrazy ? And should we drop it?

deependujha (approved):
great contribution @jwills and thanks `GPT5`. 🚀

src/litdata/streaming/reader.py
bhimrazy: yes @deependujha , I think we can drop it — limiting it to just `FileNotFoundError` should be fine here.

bhimrazy: Thanks for the PR, @jwills! 
> I’ll test this shortly.

tchaton: I think it is fine to keep

PR #705, opened 2025-09-04
[WIP] Add support for direct upload to r2 buckets

pwgardipee:
## What does this PR do?
Allows files to be directly uploaded to R2 buckets when the `output_dir` points to a lightning_storage folder.

https://www.loom.com/share/8ef98a2adc0c4f7b86543bab9679882d?sid=387f8cbd-48a2-417e-8058-2346d14736d3

**Questions for @tchaton and/or @bhimrazy:**
1. We need the `data_connection_id` in order to fetch the temporary credentials for the r2 bucket. I wasn't able to find a graceful way of threading that through from the `output_dir` to the `get_bucket_and_path` function without adding a ton of code, so I just setting an env var and reading from that. This feels very hacky so let me know if you have any recommendations.
2. It seems like we should have `get_r2_bucket_credentials` call an SDK method instead of making a direct API call. I can do this in a followup PR unless you think it's necessary now.

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/streaming/fs_provider.py
+++ b/src/litdata/streaming/fs_provider.py
@@ -223,6 +223,159 @@ class S3FsProvider(FsProvider):
 
         return not objects["KeyCount"] > 0
 
+class R2FsProvider(FsProvider):
+    def __init__(self, storage_options: Optional[dict[str, Any]] = {}):
+        super().__init__(storage_options=storage_options)
+
+        # Get data connection ID from environment variable (set by resolver)
+        data_connection_id = os.getenv("LIGHTNING_DATA_CONNECTION_ID")
+        
+        # Fetch R2 credentials from the Lightning platform and add them to the storage options
+        r2_credentials = self.get_r2_bucket_credentials(data_connection_id=data_connection_id)
+        storage_options = {**storage_options, **r2_credentials}
+        
+        self.client = S3Client(storage_options=storage_options)
+    
+    def get_r2_bucket_credentials(self, data_connection_id: str) -> dict[str, str]:
+        """
+        Fetch temporary R2 credentials for the current lightning storage connection.
+        """
+        import json
+        import requests
+        
+        try:
+            # Get Lightning Cloud API token
+            cloud_url = os.getenv("LIGHTNING_CLOUD_URL", "https://lightning.ai")
+            api_key = os.getenv("LIGHTNING_API_KEY")
+            username = os.getenv("LIGHTNING_USERNAME")
+            project_id = os.getenv("LIGHTNING_CLOUD_PROJECT_ID")
+            
+            if not all([api_key, username, project_id]):
+                raise RuntimeError("Missing required environment variables")
+                
+            # Login to get token
+            payload = {"apiKey": api_key, "username": username}
+            login_url = f"{cloud_url}/v1/auth/login"
+            response = requests.post(login_url, data=json.dumps(payload))
+            
+            if "token" not in response.json():
+                raise RuntimeError("Failed to get authentication token")
+                
+            token = response.json()["token"]
+            
+            # Get temporary bucket credentials
+            headers = {"Authorization": f"Bearer {token}", "Content-Type": "application/json"}
+            credentials_url = f"{cloud_url}/v1/projects/{project_id}/data-connections/{data_connection_id}/temp-bucket-credentials"
+            
+            credentials_response = requests.get(credentials_url, headers=headers)
+            
+            if credentials_response.status_code != 200:
+                raise RuntimeError(f"Failed to get credentials: {credentials_response.status_code}")
+                
+            temp_credentials = credentials_response.json()
+
+            endpoint_url = f"https://{temp_credentials['accountId']}.r2.cloudflarestorage.com"
+            
+            # Format credentials for S3Client
+            return {
+                "aws_access_key_id": temp_credentials["accessKeyId"],
+                "aws_secret_access_key": temp_credentials["secretAccessKey"],
+                "aws_session_token": temp_credentials["sessionToken"],
+                "endpoint_url": endpoint_url,
+                "region_name": "auto"
+            }
+            
+        except Exception as e:
+            # Fallback to hardcoded credentials if API call fails
+            print(f"Failed to get R2 credentials from API: {e}. Using fallback credentials.")
+            raise RuntimeError(f"Failed to get R2 credentials and no fallback available: {e}")
+
+    def upload_file(self, local_path: str, remote_path: str) -> None:
+        bucket_name, blob_path = get_bucket_and_path(remote_path, "r2")
+        self.client.client.upload_file(local_path, bucket_name, blob_path)
+
+    def download_file(self, remote_path: str, local_path: str) -> None:
+        bucket_name, blob_path = get_bucket_and_path(remote_path, "r2")
+        with open(local_path, "wb") as f:
+            self.client.client.download_fileobj(bucket_name, blob_path, f)
+
+    def download_directory(self, remote_path: str, local_directory_name: str) -> str:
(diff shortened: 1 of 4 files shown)


src/litdata/streaming/fs_provider.py
```diff
@@ -224,6 +224,161 @@ def is_empty(self, path: str) -> bool:
         return not objects["KeyCount"] > 0
 
 
+class R2FsProvider(FsProvider):
+    def __init__(self, storage_options: Optional[dict[str, Any]] = {}):
+        super().__init__(storage_options=storage_options)
+
+        # Get data connection ID from environment variable (set by resolver)
+        data_connection_id = os.getenv("LIGHTNING_DATA_CONNECTION_ID")
```
tchaton: Let's not do that. This won't work out of the box in Studios.

Here is how I would do it. I would pass the input_dir down to the R2 clients and using the _resolve_dir internal logic, decide what to do

src/litdata/streaming/fs_provider.py
```diff
@@ -224,6 +224,161 @@ def is_empty(self, path: str) -> bool:
...
+
+        # Fetch R2 credentials from the Lightning platform and add them to the storage options
+        r2_credentials = self.get_r2_bucket_credentials(data_connection_id=data_connection_id)
+        storage_options = {**storage_options, **r2_credentials}
+
+        self.client = S3Client(storage_options=storage_options)
```
tchaton: This won't work. The S3Client is meant to refresh the credentials. Here, it is generated once.

I would just create an R2 client instead and move the get_r2_bucket_credentials inside of it. The R2Client would expose a client property which is a refreshable r2 client e.g when the credentials have expired and the parent caller call the property, a new client is created with new credentials.

src/litdata/streaming/resolver.py
```diff
@@ -268,6 +268,9 @@ def _resolve_lightning_storage(dir_path: str) -> Dir:
     if not data_connection:
         raise ValueError(f"We didn't find any matching data connection with the provided name `{target_name}`.")
 
+    # Store data connection ID in environment variable for R2FsProvider to use
+    os.environ["LIGHTNING_DATA_CONNECTION_ID"] = data_connection[0].id
```
tchaton: I would prefer not to do that.

bhimrazy: @tchaton, I was thinking of extracting the data connection logic into a helper method, something like `_resolve_data_connection` or `_resolve_lightning_data_connection`. This method would take the `dir_path`, and just return the data connection, and also cache the result for reuse wherever needed. 

It would probably help reduce a lot of repetition within this resolver file as well.

src/litdata/streaming/fs_provider.py
```diff
@@ -224,6 +224,161 @@ def is_empty(self, path: str) -> bool:
         return not objects["KeyCount"] > 0
 
 
+class R2FsProvider(FsProvider):
```
bhimrazy: Here, I think we could inherit `S3FsProvider`.

src/litdata/streaming/resolver.py
bhimrazy: Added a PR for it here: https://github.com/Lightning-AI/litData/pull/706 — might be handy.

cc: @tchaton @pwgardipee

pwgardipee: Thanks @bhimrazy, that is very helpful - I will integrate that next.

src/litdata/streaming/fs_provider.py
pwgardipee: I tried to do this, but unfortunately now the types for `this.client` will be different (R2FsProvider has a R2Client and S3FsProvider has an S3Client) so `R2FsProvider` can't inherit from `S3FsProvider`.

src/litdata/processing/data_processor.py
```diff
@@ -233,7 +237,11 @@ def _upload_fn(
     obj = parse.urlparse(output_dir.url if output_dir.url else output_dir.path)
 
     if obj.scheme in _SUPPORTED_PROVIDERS:
-        fs_provider = _get_fs_provider(output_dir.url, storage_options)
+        # Add data connection ID to storage_options for R2 connections
+        merged_storage_options = storage_options.copy()
```
tchaton: Can we make an helper. Lot of duplicated code.

src/litdata/streaming/client.py
```diff
@@ -76,3 +76,115 @@ def client(self) -> Any:
...
+
+            token = response.json()["token"]
+
+            # Get temporary bucket credentials
+            headers = {"Authorization": f"Bearer {token}", "Content-Type": "application/json"}
+            credentials_url = (
+                f"{cloud_url}/v1/projects/{project_id}/data-connections/{data_connection_id}/temp-bucket-credentials"
+            )
+
+            credentials_response = requests.get(credentials_url, headers=headers, timeout=10)
```
tchaton: Let's add some retries, so it doesn't fail during release.

src/litdata/streaming/fs_provider.py
tchaton: Maybe the R2Client can subclass the S3client but override most of it.

tchaton: Or we can create a base class

src/litdata/streaming/client.py
```diff
@@ -11,17 +11,35 @@
...
 import botocore
+import requests
 from botocore.credentials import InstanceMetadataProvider
 from botocore.utils import InstanceMetadataFetcher
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
 
 from litdata.constants import _IS_IN_STUDIO
 
+_CONNECTION_RETRY_TOTAL = 2880
```
robTheBuildr, line 31: this number seems "magic", can we add a comment explaining or ideally link to the docs where you got this from?

src/litdata/streaming/client.py
```diff
@@ -11,17 +11,35 @@
...
+_CONNECTION_RETRY_BACKOFF_FACTOR = 0.5
```
robTheBuildr, line 33: ditto, magic number with no explanation

src/litdata/streaming/client.py
```diff
@@ -11,17 +11,35 @@
...
+_DEFAULT_REQUEST_TIMEOUT = 30  # seconds
```
robTheBuildr, line 35: ditto

src/litdata/streaming/config.py
```diff
@@ -151,6 +155,10 @@ def download_chunk_from_index(self, chunk_index: int, skip_lock: bool = False) -
 
         self.try_decompress(local_chunkpath)
 
+        from time import time
+
+        print("HERE 2", time())
```
robTheBuildr: remove debugging log

src/litdata/streaming/reader.py
```diff
@@ -455,6 +455,7 @@ def read(self, index: ChunkedIndex) -> Any:
                 try:
                     self._prepare_thread.join(timeout=_LONG_DEFAULT_TIMEOUT)
                 except Timeout:
+                    breakpoint()
```
robTheBuildr: remove debugging code

src/litdata/streaming/fs_provider.py
pwgardipee: Done ✅

src/litdata/streaming/fs_provider.py
pwgardipee: Addressed

src/litdata/streaming/fs_provider.py
pwgardipee: Addressed

PR #737, opened 2025-10-14
feat: add support to disable external version checks

### LINKED ISSUE #734: Add mechanism to disable external version checks (PyPI API calls) for air-gapped environments
## Feature

Add a mechanism to prevent LitData from making external API calls (specifically, to PyPI for version checks) in environments where egress is not permitted, such as air-gapped or highly restricted systems.

### Motivation

We are running LitData on an air-gapped system that blocks unexpected outbound network calls. Currently, LitData attempts to check the latest version by calling the PyPI API. Although the logic is safely wrapped in a try-except block, the call takes up to two minutes to time out before proceeding, which causes delays.

The first time we noticed an outbound call to pypi.org, it raised internal concerns. Upon reviewing the code, we confirmed the intent was benign, but it would be helpful to have control over whether these version checks are performed.

### Pitch

Introduce a flag or environment variable (e.g., `LITDATA_DISABLE_VERSION_CHECK=1`) to allow users to disable these external version checks and avoid unnecessary API calls and wait times in restricted environments. I do not have strong opinions on the implementation details and would welcome input from the maintainers or community.

I am happy to contribute a PR with a proposed change, but I wanted to open an issue first in case others are working on this or have thoughts on the best approach.

### Alternatives

- Continue relying on the try-except timeout. This is not ideal, as it still introduces delays.

### Additional context

The relevant code is in [`src/litdata/helpers.py`](https://github.com/Lightning-AI/litData/blob/dad316ebb955cb33a71786be9cc5b80e1f640ab5/src/litdata/helpers.py), specifically the `_get_newer_version` and `_check_version_and_prompt_upgrade` functions.

Thank you for considering this suggestion!

sanggusti:
Fixes #734

<details>
  <summary><b>Add mechanism to disable external version checks to address #734 </b></summary>

</details>

## What does this PR do?

Fixes #734 . Implemented optional environment variable `LITDATA_DISABLE_VERSION_CHECK=1` that, when set, skips the PyPI API calls entirely. This prevents delays, timeouts, and unexpected network activity in restricted environments while maintaining backward compatibility (i.e., the default behavior remains unchanged unless the flag is explicitly enabled).

## PR review

## Did you have fun?

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -33,7 +33,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - name: Install uv and setup python ${{ matrix.python-version }}
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v7
         with:
           activate-environment: true
           python-version: ${{ matrix.python-version }}
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -2,7 +2,7 @@ av >=14.0.0
 coverage ==7.10.*
 cryptography==45.0.7
 mosaicml-streaming==0.11.0
-torchaudio >=2.7.0
+torchaudio>=2.7.0,<2.9
 pytest ==8.4.*
 pytest-asyncio>=1.0.0
 pytest-cov ==7.0.0


src/litdata/helpers.py
```diff
@@ -21,13 +22,20 @@ def warn(self, message: str, stacklevel: int = 5, **kwargs: Any) -> None:
 __package_name__ = "litdata"
 
 
+def _is_version_check_disabled() -> bool:
+    """Check if external version checks are disabled via environment variable."""
+    return os.environ.get("LITDATA_DISABLE_VERSION_CHECK") == "1"
```
bhimrazy: I think it might be a good idea to move this environment variable to `constants.py` for better consistency.

You can find other constants defined here.
https://github.com/Lightning-AI/litData/blob/e69fdd0042de7f08b1b4e85b518eb950c57e79c3/src/litdata/constants.py#L53-L58

That way, we could simply reference the constant directly instead of introducing a separate function.

src/litdata/helpers.py
```diff
@@ -21,13 +22,20 @@ def warn(self, message: str, stacklevel: int = 5, **kwargs: Any) -> None:
...
+
 @functools.lru_cache(maxsize=1)
 def _get_newer_version(curr_version: str) -> Optional[str]:
     """Check PyPI for newer versions of ``litdata``.
 
     Returning the newest version if different from the current or ``None`` otherwise.
 
     """
+    if _is_version_check_disabled():
+        return None
```
bhimrazy:
```suggestion
```

I think this might not be necessary since we already handle the check in the `_check_version_and_prompt_upgrade` function.
But please feel free to correct me if I’ve missed something here.

tests/test_version_check.py
```diff
@@ -0,0 +1,132 @@
+import os
```
bhimrazy: I think we might be over-testing this a bit. We could probably reduce it to just 1–2 focused tests that cover the intended behavior with mocks.

For setting the environment variable, you can either use `patch` or `monkeypatch`, similar to these examples:

https://github.com/Lightning-AI/litData/blob/e69fdd0042de7f08b1b4e85b518eb950c57e79c3/tests/streaming/test_reader.py#L110

https://github.com/Lightning-AI/litData/blob/e69fdd0042de7f08b1b4e85b518eb950c57e79c3/tests/utilities/test_env.py#L4-L6

Also, it might be good to rename the file to `test_helpers.py` to better follow the existing file naming patterns.

src/litdata/helpers.py
sanggusti: Great catch! I've added it as constant and adjusted the code

src/litdata/helpers.py
sanggusti: Great catch! Just adjusted and passed the test. I've added it before since my test failing without it, but it was just a mock problem. Now resolved

tests/test_version_check.py
sanggusti: Have reduced the tests to only 3 of focused test on 3 scenarios:
- Verify no network call
- Default behavior then get new version
- `_check_version_and_prompt_upgrade` test

src/litdata/helpers.py
```diff
@@ -28,6 +30,8 @@ def _get_newer_version(curr_version: str) -> Optional[str]:
     Returning the newest version if different from the current or ``None`` otherwise.
 
     """
+    if _LITDATA_DISABLE_VERSION_CHECK == 1:
+        return None
     if packaging_version.parse(curr_version).is_prerelease:
         return None
```
deependujha:
```suggestion
    if _LITDATA_DISABLE_VERSION_CHECK == 1 or packaging_version.parse(curr_version).is_prerelease:
        return None
```

any issue with merging the two conditions

src/litdata/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Introduced `CHANGELOG.md` to track changes across releases ([#733](https://github.com/lightning-ai/litdata/pull/733))
+- Implement disable external version check with optional environment variable `LITDATA_DISABLE_VERSION_CHECK` ([#734](https://github.com/Lightning-AI/litData/issues/734))
```
deependujha:
```suggestion
- Add environment variable `LITDATA_DISABLE_VERSION_CHECK` to disable PyPI version check ([#734](https://github.com/Lightning-AI/litData/issues/734))
```

Maybe we could phrase the changelog entry like this for clarity.

tests/test_helper.py
```diff
@@ -0,0 +1,65 @@
...
+
+import pytest
+
+from litdata.helpers import _check_version_and_prompt_upgrade, _get_newer_version
+
+
+@pytest.fixture
+def set_env_var():
+    """Fixture to set and reset the env var."""
+    original_value = os.environ.get("LITDATA_DISABLE_VERSION_CHECK")
```
deependujha: any issue with using `monkeypatch.setenv`?

sanggusti: addressed

src/litdata/helpers.py
sanggusti: nice catch. addressed

src/litdata/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Introduced `CHANGELOG.md` to track changes across releases ([#733](https://github.com/lightning-ai/litdata/pull/733))
+- Add environment variable `LITDATA_DISABLE_VERSION_CHECK` to disable PyPI version check ([#734](https://github.com/Lightning-AI/litData/issues/734))
```
deependujha:
```suggestion
- Add environment variable `LITDATA_DISABLE_VERSION_CHECK` to disable PyPI version check ([#737](https://github.com/Lightning-AI/litData/issues/737))
```

my understanding is, we specify PR not issue.

src/litdata/constants.py
```diff
@@ -55,6 +55,7 @@
 
 _MAX_WAIT_TIME = int(os.getenv("MAX_WAIT_TIME", "120"))
 _FORCE_DOWNLOAD_TIME = int(os.getenv("FORCE_DOWNLOAD_TIME", "30"))
+_LITDATA_DISABLE_VERSION_CHECK = int(os.getenv("LITDATA_DISABLE_VERSION_CHECK", "1"))
```
deependujha:
```suggestion
_LITDATA_DISABLE_VERSION_CHECK = int(os.getenv("LITDATA_DISABLE_VERSION_CHECK", "0"))
```

By default, we should check for latest pypi version check.

tests/test_helper.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.fixture
+def set_env_var(monkeypatch):
+    """Fixture to set and reset the env var."""
+    original_value = os.environ.get("LITDATA_DISABLE_VERSION_CHECK")
+    yield
+    if original_value is None:
+        monkeypatch.delenv("LITDATA_DISABLE_VERSION_CHECK", raising=False)
+    else:
+        monkeypatch.setenv("LITDATA_DISABLE_VERSION_CHECK", original_value)
```
deependujha: I'm not sure why we need this function.

bhimrazy:
The failing test seems to be mainly due to the new versions of Torch and Torchaudio, which [[dropped](https://github.com/pytorch/audio/issues/3902#issue-3017467084)](https://github.com/pytorch/audio/issues/3902#issue-3017467084) some functionality.

Maybe we can pin the Torchaudio version for now in a separate PR and address the compatibility issues with the newer version in upcoming PRs.

tests/test_helper.py
```diff
@@ -0,0 +1,57 @@
...
+    monkeypatch.delenv("LITDATA_DISABLE_VERSION_CHECK", raising=False)
+
+    if disable_version_check is not None:
+        monkeypatch.setenv("LITDATA_DISABLE_VERSION_CHECK", disable_version_check)
+
+    # Reload both modules so constants re-evaluate environment variables
+    sys.modules.pop("litdata.constants", None)
+    sys.modules.pop("litdata.helpers", None)
+    importlib.import_module("litdata.helpers")
+    from litdata import helpers
```
bhimrazy: With the patch, I think this could be a bit simplified; otherwise, looks good.

deependujha: sure, but, test itself is very simple, so wouldn't hurt, lol.

deependujha: but let's do patch if it's redundantly complex.

tchaton (approved):

PR #740, opened 2025-10-24
Adding multisample feature along with testcases

### LINKED ISSUE #317: Add support for multi sample item in optimize and yielding from the _getitem_ of the StreamingDataset
### Motivation

It would be great to be able to create a batch of sub sample from a given sample. Right now, you can't do that.

However a user could support this.

```python

def optimize(...):

    sample = 
	return MultiSample(sample, num_samples=X)
```

Under the hood, we know this sample could be used to generate multiple random samples. 

```python

class MyStreamingDataset(StreamingDataset):

	def __getitem__(self, index, sample_id):
			sample  = super().__getitem__(index)
            
            # do some transformation
            return  data
```

A use case would be image detection where each image can be used to generate multiple sub boxes and we might want to have them as different training samples.

VijayVignesh1:
<details>

</details>

## What does this PR do?

Fixes #317

## PR review
Added support for multisample item. <br>
Basically added a sample_count parameter which creates a batch of sub samples for each sample, given a single transform function. <br>
```
Note:
Multi-sample behavior applies only when the transform is passed to the
StreamingDataset constructor (i.e., via the `transform` argument),
and not when overriding `__init__` in this subclass. 
```

Sample code:

```
    def transform_fn_sq(x, sample_idx, *args, **kwargs):
        """A simple transform function that doubles the input."""
        return x * sample_idx

    dataset = StreamingDataset(
        data_dir,
        cache_dir=str(cache_dir),
        shuffle=False,
        transform=[transform_fn_sq],
        sample_count=3,
    )
```

## Did you have fun?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 torch
-torchvision<=0.25
+torchvision
 lightning-utilities
 filelock
 numpy
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 torch
-torchvision
+torchvision<=0.25
 lightning-utilities
 filelock
 numpy
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -2,18 +2,17 @@ av >=14.0.0
 coverage ==7.10.*
 cryptography==45.0.7
 mosaicml-streaming==0.11.0
-torchaudio>=2.7.0,<2.9
 pytest ==8.4.*
 pytest-asyncio>=1.0.0
 pytest-cov ==7.0.0
 pytest-timeout ==2.4.0
-pytest-rerunfailures ==15.1
+pytest-rerunfailures ==16.0.1
 pytest-random-order ==1.2.0
 pytest-xdist >=3.8.0
 pandas
 pyarrow >=20.0.0
 polars >1.0.0
 lightning
-transformers <4.57.0
-zstd
-soundfile >=0.13.0 # required for torchaudio backend
+transformers >=4.51.0
+zstd; python_version < "3.14"
+soundfile >=0.13.0
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -16,5 +16,4 @@ lightning
 transformers >=4.51.0
 zstd; python_version < "3.14"
 soundfile >=0.13.0
-torchvision <=0.25.0 # temp pinning until https://github.com/Lightning-AI/litData/pull/802 lands
 torchcodec >=0.1.0 # check compatibility table: https://github.com/meta-pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -17,3 +17,4 @@ transformers >=4.51.0
 zstd; python_version < "3.14"
 soundfile >=0.13.0
 torchvision <=0.25.0 # temp pinning until https://github.com/Lightning-AI/litData/pull/802 lands
+torchcodec >=0.1.0 # check compatibility table: https://github.com/meta-pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -16,3 +16,4 @@ lightning
 transformers >=4.51.0
 zstd; python_version < "3.14"
 soundfile >=0.13.0
+torchvision <=0.25.0 # temp pinning until https://github.com/Lightning-AI/litData/pull/802 lands
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,12 +1,12 @@
 av >=14.0.0
-coverage ==7.10.*
+coverage ==7.13.*
 cryptography==45.0.7
-mosaicml-streaming==0.11.0
+mosaicml-streaming==0.13.0
 pytest ==8.4.*
 pytest-asyncio>=1.0.0
 pytest-cov ==7.0.0
 pytest-timeout ==2.4.0
-pytest-rerunfailures ==16.0.1
+pytest-rerunfailures ==16.1
 pytest-random-order ==1.2.0
 pytest-xdist >=3.8.0
 pandas
--- a/src/litdata/streaming/dataset.py
+++ b/src/litdata/streaming/dataset.py
@@ -62,6 +62,7 @@ class StreamingDataset(IterableDataset):
         index_path: Optional[str] = None,
         force_override_state_dict: bool = False,
         transform: Optional[Union[Callable, list[Callable]]] = None,
+        is_multisample: bool = False,
     ) -> None:
         """The streaming dataset can be used once your data have been optimised using the DatasetOptimiser class.
 
@@ -89,6 +90,7 @@ class StreamingDataset(IterableDataset):
                 If `index_path` is a full file path, it will use that directly.
             force_override_state_dict: Boolean flag for allowing local arguments to override a loaded state dict.
             transform: Optional transformation function or list of functions to apply to each item in the dataset.
+            is_multisample: If True, each index access returns multiple samples transformed by the list of functions.
         """
         _check_version_and_prompt_upgrade(__version__)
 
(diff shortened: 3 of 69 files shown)


VijayVignesh1:
@tchaton @deependujha @bhimrazy Can you verify the approach once? I can then make changes to the README.

src/litdata/streaming/dataset.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         index_path: Optional[str] = None,
         force_override_state_dict: bool = False,
         transform: Optional[Union[Callable, list[Callable]]] = None,
+        is_multisample: bool = False,
```
deependujha: how will you know how many `sample_count` user wants?
```suggestion
        sample_count: int = 1,
```

src/litdata/streaming/dataset.py
```diff
@@ -282,7 +287,8 @@ def _create_shuffler(self, cache: Cache) -> Shuffle:
         return FullShuffle(cache, seed, drop_last) if self.shuffle else NoShuffle(cache, seed, drop_last)
 
     def __len__(self) -> int:
-        return self.get_len(self.num_workers, self.batch_size if self.batch_size else 1)
+        original_len = self.get_len(self.num_workers, self.batch_size if self.batch_size else 1)
+        return original_len if not self.is_multisample else original_len * len(self.transform)
```
deependujha: This is an interesting approach, `using the number of transforms to determine dataset length implies that the user has defined as many transform functions as the number of desired samples`.

It makes sense conceptually, but I’m a bit unsure about the practicality. In most cases, the transforms won’t differ drastically; `a user could easily handle minor variations in transforms with using simple conditionals and sample_idx parameter`.

So I’m wondering if this design might add unnecessary complexity for limited flexibility.
Curious to hear your thoughts on this, @bhimrazy.

VijayVignesh1: I think that approach makes sense. Adding `samples_count` as a parameter instead of deriving it from the number of transform functions is more logical. This works well when self.transform is a single function. However, when it’s a list of transform functions, we need to decide how to handle it. In such cases, we can either override the multi-sample behavior or return `samples_count` samples for each transform function. I believe the first option is better, since the user has already defined multi-sample outputs through the list of transforms.
Thoughts? @deependujha

src/litdata/streaming/dataset.py
VijayVignesh1: This is better. I'll add this.

src/litdata/streaming/dataset.py
deependujha: @VijayVignesh1 let's wait for @bhimrazy thoughts, if he would like to suggest something.

deependujha: @VijayVignesh1 let's continue this PR.

deependujha: Also, for now, focus on one aspect of using this, like, a single `transform` method, and in subsequent PRs, handle for other cases.

src/litdata/streaming/dataset.py
```diff
@@ -202,12 +205,27 @@ def __init__(
...
             self.transform = transform
+
+        # define invalid transform conditions for multisample case
+        invalid_transform = self.sample_count > 1 and (
+            not hasattr(self, "transform")
+            or len(self.transform) > 1
+            or "sample_idx" not in signature(self.transform[0]).parameters
+        )
+        if invalid_transform:
+            logger.warning(
```
deependujha:
```suggestion
            raise ValueError(
```
I believe raising an explicit error here would be better than modifying what user wanted with(sample_count > 1).

bhimrazy:
Closing this PR due to inactivity. Please feel free to reopen or recreate it whenever convenient. 
A clearer path forward can also be finalized through further discussion over issue.

VijayVignesh1:
Hi @bhimrazy,
I believe the required feature for this PR has already been completed. Would it be possible to reopen it so we can continue the review and move it forward? If there are any outstanding concerns, I’m happy to address them.

VijayVignesh1:
@deependujha The pipeline is failing on test_wav_deserialization in VideoSerializer, even though my PR didn’t touch this file. Do you know why it might be failing?

UPDATE: The latest TorchVision release (v0.26) has deprecated `read_video`. Our checks are failing because we are using this newest version of TorchVision.
More details: [https://docs.pytorch.org/vision/0.25/io.html#video-deprecated](https://docs.pytorch.org/vision/0.25/io.html#video-deprecated)
I have pinned the torchvision version to 0.25 and below for now.

src/litdata/streaming/dataset.py
```diff
@@ -209,12 +212,27 @@ def __init__(
...
+                "Invalid transform configuration detected. "
+                "Either no transform, multiple transforms, or missing `sample_idx` parameter. "
```
deependujha: The message is ambiguous and forces users to guess the actual issue.
Prefer explicit validation with targeted errors per case, and fail fast instead of silently falling back.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 torch
-torchvision
+torchvision<=0.25
```
deependujha: let's update the tests in separate PR.

deependujha: opened #802

src/litdata/streaming/dataset.py
```diff
@@ -209,12 +212,27 @@ def __init__(
...
+            if not hasattr(self, "transform"):
+                raise ValueError("Transform is required when using sample_count > 1.")
+
+            if isinstance(self.transform, list) and len(self.transform) > 1:
+                raise ValueError("Only a single transform is allowed when using sample_count > 1.")
+
+            if isinstance(self.transform, list) and "sample_idx" not in signature(self.transform[0]).parameters:
+                raise ValueError(
+                    "The transform function must accept 'sample_idx' as a parameter when using sample_count > 1."
+                )
```
deependujha: rather than keeping in `__init__()`, we should make another helper function that does this checking

VijayVignesh1: Done

requirements/test.txt
```diff
@@ -16,4 +16,5 @@ lightning
 transformers >=4.51.0
 zstd; python_version < "3.14"
 soundfile >=0.13.0
+torchvision <=0.25.0 # temp pinning until https://github.com/Lightning-AI/litData/pull/802 lands
```
deependujha:
```suggestion
```

src/litdata/streaming/dataset.py
```diff
@@ -257,6 +264,20 @@ def set_drop_last(self, drop_last: bool) -> None:
             self.drop_last = drop_last
             self.shuffler = None  # Reset shuffler to pick up new drop_last setting
 
+    def _validate_transform_for_multisample(self) -> None:
+        """Validate the transform configuration when using `sample_count > 1`."""
+        if self.sample_count > 1:
+            if not hasattr(self, "transform"):
+                raise ValueError("Transform is required when using sample_count > 1.")
```
deependujha:
```suggestion
                raise ValueError("`transform` is required when using sample_count > 1.")
```

src/litdata/streaming/dataset.py
```diff
@@ -531,7 +566,9 @@ def __next__(self) -> Any:
 
             # `next_worker_chunks_index` is the index of the chunk that we will be working on now
             interval = self.worker_intervals[self.worker_next_chunk_index]
-            current_indexes = np.arange(interval[1], interval[2])
+
+            # current_indexes = np.arange(interval[1] * self.multisample_factor, interval[2] * self.multisample_factor)
```
deependujha:
```suggestion
```

deependujha:
Doesn't support resuming with `multisample`. Will be good to do it in the subsequent PR.

Also, `sample_count` word seems quite avg, what about `num_variants`, or something better? cc: @bhimrazy

deependujha (approved):
Thanks for the cool work. 🥳

deependujha:
requires #806

deependujha:
Closing this, as we’re not sure if this is actually very useful. Thanks for your efforts @VijayVignesh1 🙌

PR #768, opened 2025-12-08
feat: add `align_chunking` option to preserve deterministic chunk boundaries across workers

### LINKED ISSUE #758: SAME optimized dataset for DIFFERENT num_workers
## 🚀 Feature

Current the optimized dataset seems different if I'm using different `num_workers` chunk-$worker-$chunkid, can I make the final optimized always the same for different num_workers? Or can I reorganize the optimized dataset?

deependujha:
<details>

</details>

## What does this PR do?

Fixes #758

This PR introduces a new `align_chunking` option to `optimize()` and internal item-mapping logic. When enabled, items are assigned to workers using a deterministic chunk-aligned strategy that mirrors the single-worker chunk layout. All workers receive only full chunks of the specified size, except for the final worker, which may receive a partial chunk.

This eliminates the inconsistent partial-chunk patterns that occur when multiple workers independently chunk their own subsets, ensuring downstream samplers and loaders see stable chunk boundaries regardless of `num_workers`.

### Key points:

• Supports both `None` (current behavior) and an integer chunk size.
• Guarantees chunk alignment equivalent to `num_workers=1` & `num_nodes = 1`.
• Multi-node support ✅.
• Adds thorough tests verifying alignment behavior.

This provides a predictable chunk layout for users who rely on `chunk_size` semantics, especially when batch size is a multiple of chunk size, **`while maintaining full backwards compatibility`**.

---

### Code

- Optimize dataset
```python
import litdata as ld
import torch


def getter(index: int):
    return torch.Tensor(torch.full((4096, 128), index, dtype=torch.long))


if __name__ == "__main__":
    ld.optimize(
        fn=getter,
        inputs=list(range(7 * 64)),
        chunk_size=64,
        align_chunking=True,
        output_dir="worker_size_1",
        num_workers=1,
        mode="overwrite",
    )
    ld.optimize(
        fn=getter,
        inputs=list(range(7 * 64)),
        align_chunking=True,
        chunk_size=64,
        output_dir="worker_size_2",
        num_workers=2,
        mode="overwrite",
    )
```

- Stream and assert that each batch is from same chunk only.
```python
import litdata as ld

if __name__ == "__main__":
    print(" Worker num = 1 ".center(60, "-"))
    dataset = ld.StreamingDataset("worker_size_1/", shuffle=True)
    dataloader = ld.StreamingDataLoader(dataset, batch_size=64, num_workers=1)
    for i, batch in enumerate(dataloader):
        index = batch[:, 0, 0]
        min_item = index.min().item()
        max_item = index.max().item()
        assert max_item - min_item < 64, "each batch should belong to the same chunk."
        print(f"{i} ~ max-min-diff:", (index.max() - index.min()).item())


    print(" Worker num = 2 ".center(60, "-"))
    dataset = ld.StreamingDataset("worker_size_2/", shuffle=True)
    dataloader = ld.StreamingDataLoader(dataset, batch_size=64, num_workers=1)
    for i, batch in enumerate(dataloader):
        index = batch[:, 0, 0]
        min_item = index.min().item()
        max_item = index.max().item()
        assert max_item - min_item < 64, "each batch should belong to the same chunk."
        print(f"{i} ~ max-min-diff:", (index.max() - index.min()).item())
```

## PR review

## Did you have fun?

### DIFF
--- a/src/litdata/processing/data_processor.py
+++ b/src/litdata/processing/data_processor.py
@@ -23,6 +23,7 @@ import signal
 import sys
 import tempfile
 import traceback
+import warnings
 from abc import abstractmethod
 from contextlib import suppress
 from dataclasses import dataclass
@@ -319,6 +320,7 @@ def _map_items_to_workers_sequentially(
     assert isinstance(chunk_size, (int, type(None))), "chunk_size must be an integer or None"
 
     num_nodes = _get_num_nodes()
+    node_rank = _get_node_rank()
     world_size = num_nodes * num_workers
 
     if chunk_size is not None:
@@ -327,6 +329,13 @@ def _map_items_to_workers_sequentially(
         # Compute how many full chunks each worker can take
         full_chunks = len(user_items) // chunk_size
         chunks_per_worker = full_chunks // world_size
+
+        if chunks_per_worker == 0 and node_rank == 0:
+            warnings.warn(
+                f"chunk_size ({chunk_size}) is too large relative to dataset size ({len(user_items)}) "
+                f"and world_size ({world_size}). This will result in idle workers. "
+                f"Consider reducing chunk_size or using fewer workers."
+            )
 
         # Assign full chunks to all workers except the last
         num_items_per_worker = [chunks_per_worker * chunk_size for _ in range(world_size - 1)]
@@ -350,7 +359,6 @@ def _map_items_to_workers_sequentially(
     num_items_cumsum_per_worker = np.cumsum([0] + num_items_per_worker)
 
     out = []
-    node_rank = _get_node_rank()
     worker_idx_start = node_rank * num_workers
     worker_idx_end = (node_rank + 1) * num_workers
 
--- a/src/litdata/processing/functions.py
+++ b/src/litdata/processing/functions.py
@@ -431,7 +431,8 @@ def optimize(
         chunk_bytes: The maximum number of bytes to hold within a chunk.
         align_chunking: Ensures chunk boundaries match the single-worker layout by packing full chunks first
             and placing all remaining items in the final worker. Each worker will receive chunks of this size,
-            except possibly the last worker which may receive a smaller chunk.
+            except possibly the last worker which may receive a smaller chunk. Note: this will result in uneven
+            workload distribution among workers, and last worker may receive more data than others.
         compression: The compression algorithm to use over the chunks.
         encryption: The encryption algorithm to use over the chunks.
         num_workers: The number of workers to use during processing
--- a/tests/processing/test_functions.py
+++ b/tests/processing/test_functions.py
@@ -1,5 +1,6 @@
 import glob
 import io
+import math
 import os
 import random
 import shutil
@@ -104,10 +105,6 @@ def compress(index):
     return index, index**2
 
 
-def getter(index: int):
-    return torch.Tensor(torch.full((4096, 128), index, dtype=torch.long))
-
-
 def different_compress(index):
     return index, index**2, index**3
 
@@ -133,7 +130,7 @@ def test_optimize_align_chunking_requires_chunk_size(tmp_path):
 
     with pytest.raises(ValueError, match="`chunk_size` needs to be defined"):
         optimize(
-            fn=getter,
+            fn=compress,
             inputs=list(range(7 * 64)),
             chunk_bytes="1MB",
             output_dir=str(output_dir),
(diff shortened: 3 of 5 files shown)


src/litdata/processing/data_processor.py
```diff
@@ -299,24 +299,53 @@ def _upload_fn(
...
+        num_items_per_worker = [chunks_per_worker * align_chunking for _ in range(world_size - 1)]
+
+        # Last worker receives all remaining items (full chunks + optional tail)
+        remaining = len(user_items) - sum(num_items_per_worker)
+        num_items_per_worker.append(remaining)
+
+    else:
+        num_items_per_worker = len(user_items) // world_size
+
+        num_items_per_worker: list[int] = [num_items_per_worker for _ in range(world_size)]
```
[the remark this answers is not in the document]
deependujha: no intention to change original code

pwgardipee (approved):

src/litdata/processing/data_processor.py
```diff
@@ -299,24 +299,53 @@ def _upload_fn(
...
-    for worker_idx in range(len(num_items_per_worker) - 1, -1, -1):
-        if reminder == 0:
-            break
-        num_items_per_worker[worker_idx] += 1
-        reminder -= 1
+        # Compute how many full chunks each worker can take
+        full_chunks = len(user_items) // chunk_size
+        chunks_per_worker = full_chunks // world_size
+
+        # Assign full chunks to all workers except the last
```
bhimrazy, line 340: @deependujha , we have  a minor edge-case issue with worker assignment. 
In cases like:

For eg: 
```
import torch
import litdata as ld


def getter(index: int):
    return torch.Tensor(torch.full((4096, 128), index, dtype=torch.long))


if __name__ == "__main__":
    ld.optimize(
        fn=getter,
        inputs=list(range(100)),
        chunk_size=64,
        align_chunking=True,
        output_dir="test_idle_workers",
        num_workers=2,
    )
```

In this case: 
Worker 0: Gets 0 * 64 = 0 items.
Worker 1: Gets 100 - 0 = 100 items.

Although this may be a minor edge case, should we consider adding a warning when `chunks_per_worker == 0` to indicate that some workers will be idle?

src/litdata/processing/functions.py
```diff
@@ -428,6 +429,9 @@ def optimize(
         weights: Provide an associated weight to each input. This is used to balance work among workers.
         chunk_size: The maximum number of elements to hold within a chunk.
         chunk_bytes: The maximum number of bytes to hold within a chunk.
+        align_chunking: Ensures chunk boundaries match the single-worker layout by packing full chunks first
+            and placing all remaining items in the final worker. Each worker will receive chunks of this size,
+            except possibly the last worker which may receive a smaller chunk.
```
bhimrazy:
```python
align_chunking: If True, items are assigned to workers to ensure that every chunk (except the last one in the entire dataset) 
    has exactly `chunk_size` items. This mimics the chunk layout of a single-worker run. 
    Warning: This may result in uneven workload distribution or idle workers if the data size isn't large enough 
    to saturate all workers with full chunks.
```

Not necessarily this wording, but maybe a related warning could be added here.

tests/processing/test_functions.py
```diff
@@ -104,6 +104,10 @@ def compress(index):
     return index, index**2
 
 
+def getter(index: int):
+    return torch.Tensor(torch.full((4096, 128), index, dtype=torch.long))
```
bhimrazy:
```suggestion
    return index
```
I think we can simply return `index` here, or use an existing helper function if there’s one available.

tests/processing/test_functions.py
```diff
@@ -123,6 +127,54 @@ def random_image(index):
...
+            chunk_bytes="1MB",
+            output_dir=str(output_dir),
+            num_workers=1,
+            align_chunking=True,
+        )
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="too slow")
+@pytest.mark.parametrize(
+    ("num_workers", "expected_worker_chunks"),
```
bhimrazy: @deependujha , I think it might be better to define the chunk size here instead of `expected_worker_chunks`, since we can calculate and verify. We could then verify this across a few chunk sizes and input sizes, as the behavior should hold for any combination, wdyt ?

tests/streaming/test_dataloader.py
```diff
@@ -496,3 +497,33 @@ def test_dataloader_dataset_transform_inheritance(tmpdir, shuffle):
     # Verify that the transform is applied correctly
     for i, item in enumerate(complete_data):
         assert item == i * 2, f"Expected {i * 2}, got {item}"
+
+
+def getter(index: int):
+    return torch.Tensor(torch.full((4096, 128), index, dtype=torch.long))
```
bhimrazy:
```suggestion
    return torch.Tensor(torch.full((2, 2), index, dtype=torch.long))
```

Also, I think we can omit this test (i.e. `test_dataloader_with_align_chunking`), as it may not hold true for all cases. Its behavior seems to depend on the batch size, seed when shuffling is enabled. 
Alternatively, we might need to add a comment noting that the batch size should match the chunk size, or clarify the expected conditions.

src/litdata/processing/data_processor.py
deependujha: sure, adding a warning makes sense. Also, I guess, it simply means, your dataset is too small.

tests/streaming/test_dataloader.py
deependujha: adding a comment seems better option.

tests/processing/test_functions.py
deependujha: great idea. thanks.

tests/processing/test_functions.py
deependujha: thanks, i didn't give this much thought and simply went ahead with what was shared by user.

deependujha: updated.

PR #810, opened 2026-04-24
Fix row index miscalculation in ParquetLoader

### LINKED ISSUE #809: polars.exceptions.OutOfBoundsError in ParquetLoader with `low_memory=True`
## 🐛 Bug

The default `low_memory=True` parameter of `ParquetLoader` causes, `polars.exceptions.OutOfBoundsError: index N is out of bounds for sequence of length N` in some cases.

This relates to #553 

Why (from a failing example):

index_entry.chunk_size = 793
parquet_num_rows = 793
-> file-level counts match perfectly.
Row groups are:

[128, 128, 128, 128, 128, 128, 14, 11]
cumulative ends: 128, 256, 384, 512, 640, 768, 782, 793
For row_index=782:

correct mapping is last row group (size 11), local index 0
litdata instead picked previous row group (size 14) and tried local index 14 -> OOB (0..13 only)
So index 14 appears because it computed: 782 - 768 = 14 on the wrong row group.

### To Reproduce

Steps to reproduce the behavior...

<details>

```python
"""Minimal reproducible example for litdata ParquetLoader OutOfBoundsError.
Bug location: litdata/streaming/item_loader.py (~line 724)
    num_rows_per_row_group = parquet_file.metadata.row_group(0).num_rows
    row_group_index = row_index // num_rows_per_row_group
    row_index_within_group = row_index % num_rows_per_row_group
The loader reads row_group(0).num_rows once and pretends every row group has
that size. When the parquet file contains row groups of unequal sizes where
row_group(0) is larger than a later row_group, for some row_index the computed
row_index_within_group exceeds the actual number of rows in that smaller
row_group, and Polars raises:
    polars.exceptions.OutOfBoundsError: index N is out of bounds for sequence of length N
"""
from __future__ import annotations
import shutil
import tempfile
from pathlib import Path
import pyarrow as pa
import pyarrow.parquet as pq
import litdata
from litdata import StreamingDataset
from litdata.streaming.item_loader import ParquetLoader
# Row group layout: row_group(0) is larger than the rest.
# With groups [100, 50, 50, 50, 50]:
#   total = 300 rows. For row_index = 150, loader computes
#       row_group_index = 150 // 100 = 1
#       row_index_within_group = 150 % 100 = 50
#   but row_group(1) has only 50 rows (max valid index 49) → crash.
FIRST_GROUP_ROWS = 100
OTHER_GROUP_ROWS = 50
NUM_OTHER_GROUPS = 4
def write_nonuniform_parquet(path: Path) -> None:
    """Write a parquet file where row_group(0) has more rows than the rest."""
    schema = pa.schema([("value", pa.int64())])
    with pq.ParquetWriter(path, schema) as writer:
        first = pa.table({"value": list(range(FIRST_GROUP_ROWS))}, schema=schema)
        writer.write_table(first, row_group_size=FIRST_GROUP_ROWS)
        for g in range(NUM_OTHER_GROUPS):
            start = FIRST_GROUP_ROWS + g * OTHER_GROUP_ROWS
            tbl = pa.table(
                {"value": list(range(start, start + OTHER_GROUP_ROWS))},
                schema=schema,
            )
            writer.write_table(tbl, row_group_size=OTHER_GROUP_ROWS)
def main() -> None:
    tmp = Path(tempfile.mkdtemp(prefix="litdata_bug_"))
    print(f"workdir: {tmp}")
    try:
        data_dir = tmp / "parquet_data"
        data_dir.mkdir()
        write_nonuniform_parquet(data_dir / "chunk_0000.parquet")
        pf = pq.ParquetFile(data_dir / "chunk_0000.parquet")
        sizes = [pf.metadata.row_group(i).num_rows for i in range(pf.num_row_groups)]
        print(f"row_group sizes: {sizes}  total rows: {sum(sizes)}")
        litdata.index_parquet_dataset(str(data_dir))
        dataset = StreamingDataset(
            input_dir=str(data_dir),
            cache_dir=str(tmp / "cache"),
            item_loader=ParquetLoader(low_memory=True),
            shuffle=False,
            drop_last=False,
        )
        print(f"dataset length: {len(dataset)}")
        for i, item in enumerate(dataset):
            if i % 25 == 0:
                print(f"  row {i}: {item}")
    finally:
        shutil.rmtree(tmp, ignore_errors=True)
if __name__ == "__main__":
    main()

```

</details>

Alternatively, you can share a fully reproducible [Lightning Studio](https://lightning.ai/studios) environment:

> A simple guide on how to create such a studio can be found [here](https://www.youtube.com/watch?v=YcW-2Zt_bFg&ab_channel=LightningAI).

1. Create a [Studio](https://lightning.ai/studios).
2. Reproduce the issue in the Studio.
3. [Publish the Studio](https://lightning.ai/docs/overview/studios/publishing#how-to-publish).
4. Paste the Studio link here.

### Additional context

<details>
  <summary>Environment detail</summary>

- PyTorch Version (e.g., 1.0):
- OS (e.g., Linux):
- How you installed PyTorch (`conda`, `pip`, source):
- Build command you used (if compiling from source):
- Python version:
- CUDA/cuDNN version:
- GPU models and configuration:
- Any other relevant information:

</details>

vini-fda:
<details>

</details>

## What does this PR do?

Fixes https://github.com/Lightning-AI/litData/issues/809.

This replaces the uniform-size arithmetic with a per-chunk prefix-sum of row-group sizes, computed once when the ParquetFile is first opened, then use `bisect.bisect_right(offsets, row_index) - 1` to locate the group and `row_index - offsets[group]` for the offset inside it. 

The same `num_rows_per_row_group` value is also used in the cache-eviction check at  https://github.com/Lightning-AI/litData/blob/1fdfad72d92145305d2272483ef0be6234944cac/src/litdata/streaming/item_loader.py#L749. 

But that check needs the actual size of the current group `(offsets[g+1] - offsets[g])`, otherwise with uneven groups memory either leaks (never hits threshold) or is freed too early (forcing re-reads).

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -62,7 +62,7 @@ jobs:
           pytest tests \
               --ignore=tests/processing \
               --ignore=tests/raw \
-              -n 2 --cov=litdata --durations=0 --timeout=120 --capture=no --verbose
+              -n 2 --dist=loadgroup --cov=litdata --durations=0 --timeout=120 --capture=no --verbose
 
       - name: Run processing tests sequentially
         run: |
--- a/src/litdata/streaming/item_loader.py
+++ b/src/litdata/streaming/item_loader.py
@@ -642,6 +642,7 @@ class ParquetLoader(BaseItemLoader):
         self._df: dict[int, Any] = {}
         self._chunk_row_groups: dict[int, Any] = {}
         self._chunk_row_group_item_read_count: dict[int, Any] = {}
+        self._chunk_row_group_offsets: dict[int, list[int]] = {}
 
     def generate_intervals(self) -> list[Interval]:
         intervals = []
@@ -712,18 +713,28 @@ class ParquetLoader(BaseItemLoader):
         Returns:
             Any: The dataframe row corresponding to the specified index.
         """
+        import bisect
+
         import polars as pl
         import pyarrow.parquet as pq
 
         # Load the Parquet file metadata if not already loaded
         if chunk_index not in self._df:
-            self._df[chunk_index] = pq.ParquetFile(chunk_filepath)
+            parquet_file = pq.ParquetFile(chunk_filepath)
+            self._df[chunk_index] = parquet_file
+            # Precompute cumulative row offsets as a prefix-sum so lookup works for row groups of any size.
+            offsets = [0]
+            num_row_groups = parquet_file.metadata.num_row_groups
+            for i in range(num_row_groups):
+                num_rows = parquet_file.metadata.row_group(i).num_rows
+                offsets.append(offsets[-1] + num_rows)
+            self._chunk_row_group_offsets[chunk_index] = offsets
 
-        # Determine the row group and the row index within the row group
-        parquet_file = self._df[chunk_index]
-        num_rows_per_row_group = parquet_file.metadata.row_group(0).num_rows
-        row_group_index = row_index // num_rows_per_row_group
-        row_index_within_group = row_index % num_rows_per_row_group
+        # Locate the row group containing row_index and the offset inside it.
+        offsets = self._chunk_row_group_offsets[chunk_index]
+        row_group_index = bisect.bisect_right(offsets, row_index) - 1
+        row_index_within_group = row_index - offsets[row_group_index]
+        row_group_size = offsets[row_group_index + 1] - offsets[row_group_index]
 
         # Check if the row group is already loaded
         if chunk_index in self._chunk_row_groups and row_group_index in self._chunk_row_groups[chunk_index]:
@@ -746,7 +757,7 @@ class ParquetLoader(BaseItemLoader):
 
         # Check if the row group has been fully read and release memory if necessary
         read_count = self._chunk_row_group_item_read_count[chunk_index][row_group_index]
-        if read_count >= num_rows_per_row_group:
+        if read_count >= row_group_size:
             # Release memory for the fully read row group
             del self._chunk_row_groups[chunk_index][row_group_index]
             del self._chunk_row_group_item_read_count[chunk_index][row_group_index]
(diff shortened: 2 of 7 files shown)


Borda (approved):

vini-fda:
@Borda I had to fix three unrelated issues that were causing CI problems:

1. Example on macos-14, python 3.11: https://github.com/Lightning-AI/litData/actions/runs/26457907073/job/77897116939?pr=810

   Tests using `clean_pq_index_cache` all share `~/.lightning/chunks` and the fixture `shutil.rmtree`s it on setup/teardown. Under `pytest -n 2` two of them landed on different xdist workers and deleted each other's chunks mid-iteration. Fixed in bac1292 by auto-tagging those tests with `xdist_group("hf_default_cache")` and adding `--dist=loadgroup` to the CI invocation so they serialize onto one worker. The first attempt still scattered them across workers because my `pytest_collection_modifyitems` hook ran after xdist's own hook had already finalized nodeids; 8e3ffed adds `@pytest.hookimpl(tryfirst=True)` so the marker is in place before xdist appends the `@group` suffix that `LoadGroupScheduling` uses for routing.

2. Example on windows-2022, python 3.11: https://github.com/Lightning-AI/litData/actions/runs/26457907073/job/77897116885?pr=810

   `get_parquet_indexer_cls` runs `urlparse` on the input path. On Windows a path like `C:\Users\...` parses with `scheme='c'`, so it failed the `scheme in ("local", "")` check and raised `ValueError`. Fixed in bcb13c9 by treating any single-letter alphabetic scheme as a Windows drive letter and dispatching to `LocalParquetDir`.
 
 3. Example on ubuntu-22.04, python 3.11: `optimize()` writes intermediate chunks to `tempfile.gettempdir() + "/chunks"` (i.e. `/tmp/chunks` on Linux), configurable via `DATA_OPTIMIZER_CACHE_FOLDER`. Two streaming tests calling `optimize()` on different xdist workers both wrote to the same `/tmp/chunks` and the upload worker raced its sibling's cleanup, producing `FileNotFoundError: '/tmp/chunks/chunk-0-0.bin'`. Every `optimize()`-using test in `tests/processing/test_data_processor.py` already monkeypatches that env var to a per-test tmp dir, and the two streaming tests just didn't. Fixed in ea2d8b8 by adding the same monkeypatch.

Borda:
> Borda I had to fix three unrelated issues that were causing CI problems

Thank you for reaching out, however you shall ping maintainers ⚡

vinicius-freitas-nubank:
@tchaton @justusschock Any feedback here?

tchaton (approved):

PR #829, opened 2026-05-27
fix: natural-sort chunks in train_test_split to fix ordering with 10+ chunks

### LINKED ISSUE #826: The train_test_split utility doesn't respect original order when shuffle=False
## 🐛 Bug
The train_test_split code in litdata/utilities/train_test_split.py seems to rely on the alphabetical filename ordering of chunks, which is unreliable.
Stepping through the code I noticed the very first split contains files
curr_chunk_filename=['chunk-0-0.bin', 'chunk-1-0.bin', 'chunk-10-0.bin', 'chunk-2-0.bin', 'chunk-3-0.bin', 'chunk-4-0.bin', 'chunk-5-0.bin', 'chunk-6-0.bin']
on L91
Notice that chunk 10 snuck in there.

From testing it seems there are 2 failure paths:
- more than 10 chunks are written per worker
- more than 10 workers are used
As soon as a chunk has the name `chunk-10-x` or `chunk-x-10` it no longer sorts properly.

This is especially relevant when working with temporal data.

MohammadYusif:
## What

When a dataset has 11 or more chunks, `train_test_split` was ordering chunks by lexicographic filename sort, placing `chunk-10-0.bin` between `chunk-1-0.bin` and `chunk-2-0.bin`. This means the first split could receive `chunk-10` before `chunk-2` — silently corrupting temporal or ordered datasets.

Closes #826

## Root cause

`train_test_split` re-orders chunks by filtering `original_chunks` (from `index.json`) using a set of subsampled filenames. `index.json` itself is written with chunks in lexicographic order when more than 9 chunks exist, so the ordering passed to `subsample_filenames_and_roi` is wrong.

## Fix

Apply a natural sort (numeric-aware) to `subsampled_chunks` and the paired `region_of_interest` list immediately after the filter step, before any shuffle is applied.

```python
paired = sorted(
    zip(subsampled_chunks, dummy_subsampled_roi),
    key=lambda p: [int(x) if x.isdigit() else x
                   for x in re.split(r"(\d+)", p[0]["filename"])],
)
```

When `shuffle=True` the subsequent shuffle randomises order anyway; when `shuffle=False` the caller now receives chunks in true numeric sequence.

## Test

Added `test_train_test_split_natural_sort_ordering` in `tests/utilities/test_train_test_split.py`:

- Creates 11 chunks (22 items, `chunk_size=2`)
- Asserts each split's files are internally natural-sorted
- Asserts splits are non-overlapping and together cover all chunks
- Asserts `chunk-10-0.bin` is not placed before `chunk-9-0.bin`

## Notes

PR #828 fixes natural sort in the writer merge step for new datasets. This fix is complementary and handles existing datasets whose `index.json` was already written in lexicographic order.

### DIFF
--- a/tests/utilities/test_train_test_split.py
+++ b/tests/utilities/test_train_test_split.py
@@ -167,20 +167,22 @@ def test_train_test_split_natural_sort_ordering(tmpdir):
     train_files = train_ds.subsampled_files
     test_files = test_ds.subsampled_files
 
-    import re
-
-    def _natural_key(name: str) -> list:
-        return [int(x) if x.isdigit() else x for x in re.split(r"(\d+)", name)]
-
-    # Files within each split must be in natural order.
-    assert train_files == sorted(train_files, key=_natural_key)
-    assert test_files == sorted(test_files, key=_natural_key)
-
-    # The two splits must be non-overlapping and together cover all chunks.
-    all_files = train_files + test_files
-    assert len(all_files) == len(set(all_files)), "splits overlap"
-    assert sorted(all_files, key=_natural_key) == sorted(dataset.subsampled_files, key=_natural_key)
-
-    # chunk-10 must not appear in the training split when smaller-indexed chunks
-    # (chunk-2 … chunk-10) are still available — i.e. it must come after chunk-9.
-    assert "chunk-10-0.bin" not in train_files or "chunk-9-0.bin" in train_files
+    # With natural sort ordering, chunks 0-5 go to train and chunks 5-10 go to test.
+    # chunk-10 must appear last, not between chunk-1 and chunk-2 as lexicographic
+    # sort would place it.
+    assert train_files == [
+        "chunk-0-0.bin",
+        "chunk-1-0.bin",
+        "chunk-2-0.bin",
+        "chunk-3-0.bin",
+        "chunk-4-0.bin",
+        "chunk-5-0.bin",
+    ]
+    assert test_files == [
+        "chunk-5-0.bin",
+        "chunk-6-0.bin",
+        "chunk-7-0.bin",
+        "chunk-8-0.bin",
+        "chunk-9-0.bin",
+        "chunk-10-0.bin",
+    ]
--- a/tests/utilities/test_train_test_split.py
+++ b/tests/utilities/test_train_test_split.py
@@ -1,10 +1,12 @@
+import json
+import os
 import platform
 import sys
 
 import pytest
 
 from litdata import StreamingDataLoader, StreamingDataset, train_test_split
-from litdata.constants import _ZSTD_AVAILABLE
+from litdata.constants import _INDEX_FILENAME, _ZSTD_AVAILABLE
 from litdata.streaming.cache import Cache
 
 IS_WINDOWS = sys.platform.startswith("win") or platform.system() == "Windows"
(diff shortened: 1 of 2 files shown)


tchaton (approved):

tests/utilities/test_train_test_split.py
```diff
@@ -148,3 +148,39 @@ def test_train_test_split_with_shuffle_parameter(tmpdir, compression):
...
+        return [int(x) if x.isdigit() else x for x in re.split(r"(\d+)", name)]
+
+    # Files within each split must be in natural order.
+    assert train_files == sorted(train_files, key=_natural_key)
+    assert test_files == sorted(test_files, key=_natural_key)
+
+    # The two splits must be non-overlapping and together cover all chunks.
+    all_files = train_files + test_files
+    assert len(all_files) == len(set(all_files)), "splits overlap"
+    assert sorted(all_files, key=_natural_key) == sorted(dataset.subsampled_files, key=_natural_key)
```
deependujha: can you please simplify it? It seems complex for no reason. We can simply create an array of expected filenames and assert if they are equal.

tests/utilities/test_train_test_split.py
```diff
@@ -148,3 +148,39 @@ def test_train_test_split_with_shuffle_parameter(tmpdir, compression):
...
+
+    # chunk-10 must not appear in the training split when smaller-indexed chunks
+    # (chunk-2 … chunk-10) are still available — i.e. it must come after chunk-9.
+    assert "chunk-10-0.bin" not in train_files or "chunk-9-0.bin" in train_files
```
deependujha: do you mean `and`?

deependujha (commented):
Thanks for the fix. Can you please simplify the tests!

tests/utilities/test_train_test_split.py
MohammadYusif: The or was intentional

it's a logical implication: if chunk-10 is in train, then chunk-9 must also be in train (i.e. chunk-10 can't appear before chunk-9). But it's moot now since I've replaced the whole block with explicit expected arrays.

tests/utilities/test_train_test_split.py
MohammadYusif: Done.

replaced the _natural_key helper and all the indirect sort comparisons with direct equality checks against the expected filename lists. With 22 items at chunk_size=2 and splits=[0.5, 0.5] the split is deterministic, so hardcoding the arrays makes the intent immediately obvious.

tests/utilities/test_train_test_split.py
```diff
@@ -148,3 +148,41 @@ def test_train_test_split_with_shuffle_parameter(tmpdir, compression):
     assert shuffled_combined != no_shuffle_combined
 
     assert no_shuffle_combined == my_streaming_dataset.subsampled_files
+
+
+def test_train_test_split_natural_sort_ordering(tmpdir):
```
deependujha, line 155: Hi @MohammadYusif, expectation should be that this test fails without your changes, but it still passes.

refer tot the comment here: https://github.com/Lightning-AI/litData/issues/826#issuecomment-4564889745

deependujha (changes requested):
This is not a `train_test_split` issue, but rather `writers` bug when `num_workers >=10`. Correct fix is in: #828

tests/utilities/test_train_test_split.py
MohammadYusif: Good catch — the previous test used Cache with a single writer, which writes index.json already in natural order, so the fix never actually did anything in that test.

The real bug happens during the merge step with 10+ workers: sorted(index_files) in writer.py sorts per-worker files lexicographically, placing chunk-0-10.bin between chunk-0-1.bin and chunk-0-2.bin in the merged index.json.

Updated the test to reproduce this: after building the cache, it rewrites index.json with a lexicographic sort (mimicking exactly what the multi-worker merge produces), then asserts train_test_split still returns chunks in natural order. This test now fails without the fix.

MohammadYusif:
Closing this in favour of #828, which correctly fixes the root cause in `writer.py` — the `sorted(index_files)` call during the multi-worker merge. Fixing it at the writer level is the right approach; patching the read side in `train_test_split` was addressing the symptom in the wrong layer.

deependujha:
Thank you @MohammadYusif for the efforts and your time. Would love more contributions from you in [lightning ai](https://lightning.ai/) ecosystem. 🚀⚡️

PR #842, opened 2026-07-14
ci: [experiment] force torch >=2.13 to confirm CI hang

bhimrazy:
## What does this PR do?

Control experiment (do not merge). Forces `torch >=2.13` to confirm it causes the processing-test CI hang. Pairs with #841 (`torch <2.13`, expected green).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-torch
+torch >=2.13.0  # experiment: force torch 2.13 to confirm CI hang
 torchvision
 lightning-utilities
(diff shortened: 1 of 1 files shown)


bhimrazy:
Closing this experiment — we ruled out torch as the cause of the CI failures. Root cause is pyarrow 25.0.0, which segfaults during parquet indexing (16 failing tests in tests/streaming/test_parquet.py). The real fix is in #841, which now pins pyarrow <25 instead.

PR #852, opened 2026-07-24
perf(streaming): async S3 chunk prefetch, cache eviction, obstore downloads

tchaton:
## Summary

- **Default remote path:** process workers + async chunk prefetch (obstore). Opt out with `LITDATA_ASYNC_CHUNK_PREFETCH=0`.
- Async prepare thread: stream to disk (`adownload_file`), reuse event loop, non-blocking batch drain, floor `max_pre_download` to 4 for gather width.
- Sync S3 downloads also prefer obstore (boto3 fallback). Stream chunk size override: `LITDATA_OBSTORE_STREAM_MIN_CHUNK_MIB` (default 8).
- **Cache eviction:** size-first deletes, shared `.litdata_cache_slots` for multi-worker budgets, force-redownload keeps slots; drain all pending deletes on `END_TOKEN` so tiny caches fully clear.
- JPEG deserialize uses `ImageReadMode.RGB`.
- Removed experimental `use_threading` (real S3 ~2.3× slower than process workers).
- Benches under `scripts/bench/` (including `bench_obstore_*.py`, `bench_s3_full_epochs.py`).

## Real-S3 ImageNet (this Studio)

Dataset: `lightning_data_search`, 48 workers, batch 256, wipe `/cache/chunks`, `PYTHON_GIL=0`, transforms `RandomResizedCrop(224)+flip+fp16`.

| Mode | ep0 | ep1 | mean |
|------|-----|-----|------|
| process + sync (boto3) | 10525 | 12161 | 11284 |
| **process + async (`max_pre=4`)** | **10836** | **12154** | **11457** |
| process + async (`max_pre=8`) | 10554 | 12266 | 11345 |

Warm epoch ~12.2k is decode/transform-bound. Prefer `max_pre_download=4`.

### Cache budget

Warn below **25GB**. Slot gate only for realistic budgets (≥10MB + delete-when-processed). Real S3 check at 10GB / 48w: peak ~1.07× limit.

### Obstore benches

```bash
PYTHON_GIL=0 python -Xgil=0 scripts/bench/bench_obstore_vs_boto3.py
PYTHON_GIL=0 python -Xgil=0 scripts/bench/bench_obstore_imagenet_grid.py --workers 48 --chunk-sizes 2,4,8,16
PYTHON_GIL=0 python -Xgil=0 scripts/bench/bench_s3_full_epochs.py --label run --epochs 2 --workers 48 --batch-size 256 --max-pre-download 4
```

## Test plan

- [x] `tests/streaming/test_cache_eviction.py`
- [x] `tests/streaming/test_async_prefetch.py`
- [x] `tests/streaming/test_next_streaming_optimizations.py`
- [x] lock cleanup / download-reader overlap / JPEG writer updates
- [ ] CI


tchaton:
Follow-up: `use_threading=True` now requires a free-threaded (no-GIL) Python runtime (`not sys._is_gil_enabled()`). Regular GIL builds raise a clear `RuntimeError`.

PR #856, opened 2026-07-25
chore(tests): speed up streaming pytest suite

tchaton:
## Summary
Speed up the streaming pytest suite **without cutting resume/MP regression coverage**.

**Safe speedups (kept):**
- Reuse shared Cache dirs per `(len1, len2)` inside each xdist worker
- Shrink dataloader fixture sizes where assertions still exercise restore/epochs
- Replace fixed `sleep(1/2)` waits with polls / backdated clocks
- Raise streaming xdist to `-n 4` on Linux only (macOS/Windows stay at 2)

**Coverage (restored):**
- Full `num_workers=[0,2]` grids for RNG + dataloader state tests
- Full `length × resume × shuffle` matrices for partial/complete resume stay on **`num_workers=2` every cell**
- Dataset-only resume tests keep `num_workers=0` (no DataLoader workers are used there)

## Test plan
- [x] Local smoke of multi-worker resume cells
- [ ] CI: ubuntu streaming step still green; wall-time improvement from shared caches / xdist / fixture size only

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -61,10 +61,17 @@ jobs:
 
       - name: Run fast tests in parallel
         run: |
+          # Ubuntu runners can take more xdist workers; keep macOS/Windows at 2 to
+          # avoid spawn/resource pressure from DataLoader multiprocessing tests.
+          if [ "${{ runner.os }}" = "Linux" ]; then
+            XDIST_WORKERS=4
+          else
+            XDIST_WORKERS=2
+          fi
           pytest tests \
               --ignore=tests/processing \
               --ignore=tests/raw \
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-testing.yml
```diff
@@ -61,10 +61,17 @@ jobs:
...
+          if [ "${{ runner.os }}" = "Linux" ]; then
+            XDIST_WORKERS=4
+          else
+            XDIST_WORKERS=2
+          fi
           pytest tests \
               --ignore=tests/processing \
               --ignore=tests/raw \
-              -n 2 --dist=loadgroup --cov=litdata --durations=0 --timeout=120 --capture=no --verbose
+              -n "${XDIST_WORKERS}" --dist=loadgroup --cov=litdata --durations=0 --timeout=120 --capture=no --verbose
```
deependujha, line 74: we can also try `standalone script` for running tests

PR #866, opened 2026-08-09
Fix multi-node append duplicating existing chunks in index.json

### LINKED ISSUE #865: Multi-node optimize(mode="append") duplicates existing dataset chunks N times in index.json
#### Describe the bug

When calling `optimize(..., mode="append")` with `num_nodes > 1`, pre-existing dataset chunk metadata is included in every node's intermediate `{node_rank}-index.json` file. When the master node merges these intermediate indices, the existing chunks are concatenated repeatedly, duplicating them N times in the final `index.json`, where N is the number of nodes.

```text
Intended:

Existing [A, B]
      ├──► Node 0 [C] → 0-index.json ──┐
      └──► Node 1 [D] → 1-index.json ──┴──► Merge → [A, B, C, D]


Current Bug:

Existing [A, B]
      ├──► Node 0 [A, B, C] → 0-index.json ──┐
      └──► Node 1 [A, B, D] → 1-index.json ──┴──► Merge
                                                   ↓
                                            [A, B, C, A, B, D]
```

#### Root Cause

1. **`data_processor.py`**: Every node passes `existing_index` to `_merge_no_wait()`.

2. **`writer.py`**: `_merge_no_wait()` adds `existing_index["chunks"]` to each node's local index.

3. **`data_processor.py`**: The master node concatenates all node index files during the final merge, causing the existing chunks to be repeated once per node.

#### Suggested Fix

The existing index should be excluded from intermediate node-level merges and added only once during the final global merge.

In **`DataProcessor._done()`**:

```python
existing_index = getattr(self, "existing_index", None)

merge_cache._merge_no_wait(
    node_rank if num_nodes > 1 else None,
    None if num_nodes > 1 else existing_index,
)
```

In **`DataProcessor._upload_index()`**:

```python
merge_cache._merge_no_wait(
    existing_index=getattr(self, "existing_index", None),
)
```

### Expected behavior

The final `index.json` should contain pre-existing chunks exactly once, followed by the newly generated chunks from all nodes:

```text
[A, B, C, D]
```

rather than:

```text
[A, B, C, A, B, D]
```

eeshsaxena:
Fixes #865.

When you run `optimize(mode="append")` across more than one node, the existing dataset chunks end up repeated once per node in the final `index.json`.

The cause is in the two-stage merge. Each node calls `_merge_no_wait(node_rank, existing_index)` in `_done`, and `_merge_no_wait` prepends `existing_index["chunks"]` to that node's `{node_rank}-index.json`. So every node's index file already contains the existing chunks. The last node then merges all of those per-node files together, and the existing chunks come along once for each node:

```
existing [A, B], node 0 adds [C], node 1 adds [D]
  node 0 -> 0-index.json = [A, B, C]
  node 1 -> 1-index.json = [A, B, D]
  final merge -> [A, B, C, A, B, D]   # wanted [A, B, C, D]
```

The fix adds the existing index only once. The single-node path is unchanged: it still folds the existing index in at the node-level merge. For multiple nodes, the per-node merges now carry only their own new chunks, and the existing index is folded in at the final cross-node merge instead:

```
  node 0 -> 0-index.json = [C]
  node 1 -> 1-index.json = [D]
  final merge (with existing) -> [A, B, C, D]
```

I added a test in `tests/streaming/test_writer.py` that walks both patterns: it shows the old per-node folding produces `[A, B, C, A, B, D]`, and the new final-merge folding produces `[A, B, C, D]`. It passes locally.


tchaton (approved):
Thanks @eeshsaxena

PR #871, opened 2026-08-10
feat(raw): make dataset concurrency budget dynamic with process-local…

### LINKED ISSUE #870: Concurrency Budget is Static & Estimated From Hardcoded Magic Constants
## Problem

In `StreamingRawDataset`, the adaptive concurrency budget for remote downloads was previously calculated using hardcoded static assumptions:

```python
_ASSUMED_AGGREGATE_BANDWIDTH_BPS = 100 * 1024 * 1024  # ~100 MB/s
_ASSUMED_REQUEST_RATE = 6000.0                         # target req/s
_ASSUMED_REQUEST_LATENCY_S = 0.040                     # 40 ms RTT
_DEFAULT_MEDIAN_FILE_BYTES = 256 * 1024
```

These assumptions do not reflect the large variation in real-world cloud environments.

* **Low-Bandwidth / Shared Nodes:** Multi-tenant EKS/Kubernetes nodes may provide only ~10–20 MB/s of effective bandwidth. The static model can overestimate sustainable concurrency, leading to **HTTP 429 throttling, socket saturation, and connection timeouts**.

* **High-Bandwidth / High-Performance Storage:** Dedicated A100/H100 nodes with multi-Gbps networking or low-latency object storage can support significantly higher throughput. The conservative static cap can unnecessarily limit concurrency, resulting in **GPU underutilization and reduced data-loading throughput**.

* **No Empirical Feedback Loop:** Actual download throughput and request latency were not incorporated into subsequent concurrency decisions. As a result, the concurrency budget remained largely independent of the performance characteristics of the current network and remote storage.

### in short:

The previous concurrency model was therefore **configuration-driven rather than measurement-driven**. A more robust approach should use empirical runtime measurements—such as observed download throughput and request latency—to dynamically estimate sustainable concurrency for the current execution environment.

hillhack:
fix: https://github.com/Lightning-AI/litData/issues/870
### 1. Added Process-Local EMA Tracker

Implemented a thread-safe, pickleable `BandwidthTracker` in `raw/dataset.py`.

* Tracks **throughput** for large downloads (`≥ 1 MiB`).
* Tracks **request latency** for small GET requests (`< 256 KB`).
* Uses EMA with `α = 0.2`.
* Separates observations by payload size to prevent small requests from skewing throughput and large downloads from skewing latency.

### 2. Added Empirical Warm-up

Added `_WARMUP_SAMPLE_THRESHOLD = 5`.

* Uses conservative defaults until 5 observations are collected.
* After warm-up, concurrency is driven by measured network performance.

### 3. Added Download Telemetry

Instrumented all `CacheManager` download paths:

* `adownload_file`
* `_fetch_bytes`
* `_fetch_ranges`
* `_hedged`

Every completed download now feeds its observed latency and payload size back into `BandwidthTracker`.

### 4. Dynamic Concurrency Budget

Updated `_aggregate_concurrency_budget` to use measured bandwidth and latency:

$$
C = \max\left(
\frac{\text{Bandwidth}}{\text{Median File Size}},
\text{Bandwidth} \times \text{Latency}
\right)
$$

The result is clamped between **32 and 512**.

This allows concurrency to automatically scale **down on constrained networks** and **up on high-performance networks**.

### 5. Added Environment Overrides

Added configurable environment variables:

* `LITDATA_ASSUMED_BANDWIDTH_BPS`
* `LITDATA_ASSUMED_REQUEST_LATENCY_S`
* `LITDATA_AGGREGATE_CONCURRENCY_BUDGET_CAP`
* `LITDATA_AGGREGATE_CONCURRENCY_BUDGET_FLOOR`
* `LITDATA_SINGLE_PROCESS_CONCURRENCY_CAP`

This provides administrators with manual control without modifying the code.

### 6. Added Unit Tests

Added coverage in `tests/raw/test_dataset.py` for:

* EMA calculation and size partitioning
* `BandwidthTracker` pickling
* Warm-up gating
* High/low bandwidth adaptation
* Environment variable overrides

## How This Fixes the Problem

**Self-tuning:** Replaces static network assumptions with empirical measurements.

**Slow/shared networks:** Lower measured bandwidth automatically reduces concurrency, helping avoid socket exhaustion, timeouts, and HTTP 429 throttling.

**Fast networks:** Higher measured bandwidth increases concurrency (up to 512), improving NIC utilization and reducing GPU data starvation.

**Result:** `StreamingRawDataset` now adapts its download concurrency to the actual runtime network and storage characteristics.


tchaton (changes requested):
## Review

Direction is right (measure, then adapt), but I would not merge as-is. The feedback loop can raise concurrency for the wrong reason, cannot actually drop below today’s floor, and the ImageNet-sized path never measures bandwidth.

### What works

- Process-local, pickleable tracker with a lock and `__getstate__` / `__setstate__` fits DataLoader fork/spawn.
- Splitting “small GET → latency” vs “large GET → bandwidth” is the right idea.
- Warm-up before swapping defaults, plus env overrides, is a reasonable control surface.
- Tests cover EMA math, pickling, the 5-sample gate, and a few env knobs.

### Blocking: the model does not do what the PR / #870 claim

**1. ImageNet-sized files never update bandwidth.**

Latency observations stop at 256 KiB; bandwidth starts at 1 MiB. Typical ~200 KB JPEGs only move `request_latency_s_ema`. After warm-up, `obs_bps` stays `None` and the code still uses the hardcoded 100 MB/s.

The PR formula

```text
C = max(Bandwidth / Median File Size, Bandwidth × Latency)
```

is also not what the code does. It is still `(bps × 0.5s) // median` vs `assumed_rate × latency` (`LITDATA_ASSUMED_REQUEST_RATE` / 6000 is unchanged).

**2. Small-GET “latency” includes transfer time, so slowness increases concurrency.**

`_fetch_bytes` records full wall time. Little’s-law is `6000 * obs_lat` (rate is still assumed). A slow 200 KB GET looks like a large RTT → budget walks toward 512. That is the opposite of the 429 / socket-saturation story in #870.

That also explains the posted bench better than “EMA bandwidth”:

| workers | static budget 256 | if latency-inflated → 512 |
|---|---|---|
| w=8 | 32 permits/worker | 64 — already saturated, ~0% |
| w=24 | `max(8, 256//24) = 10` | `512//24 = 21` — matches **+58%** |

So the win is likely “more slots at high `num_workers`”, not measured NIC share. Worth a before/after of **permit counts**, not only ips.

**3. The floor makes “scale down on slow networks” false.**

`test_concurrency_budget_high_and_low_bandwidth_adaptation` asserts the 5s / 10 MiB case still budgets **32**. Realized aggregate is also `max(budget, 8 × num_workers)`, so w=24 is already 192 in-flight before any EMA. Constrained EKS will not get fewer sockets from this PR.

**4. Measurement quality is weak on the instrumented paths.**

- Ranged `gather`: wall time of **parallel** ranges over **full** size → bandwidth overstated.
- Hedged GET: clock includes hedge wait / winner, not one transfer.
- Large-object bps subtracts latency EMA (or the 40 ms default) with `max(0.001, dur - lat)`. A fast 1 MiB GET becomes ~1 GB/s and the EMA spikes.
- Mid-size objects (256 KiB–1 MiB) increment `sample_count` but update neither EMA, then unlock “empirical” mode with no empirical bps.

### Other issues

- **`_cached_permits` is no longer “once per process”.** Invalidating on sample 5 and every 10th sample is fine, but `_effective_download_permits`’s docstring is now wrong. Replacing the `asyncio.Semaphore` does not shrink in-flight work already holding the old one; you can overshoot the new budget.
- **`sample_count` is read unlocked** in `_record_download_observation` (`prev_count = self._bandwidth_tracker.sample_count`). Use the return of `get_metrics()` / the lock.
- **`_WARMUP_SAMPLE_THRESHOLD = 5` is in the PR text only.** The `5` and `% 10` are magic numbers in three places.
- **Drive-by:** `litdata-raw-pool` → `asyncio_litdata-raw-pool` is unrelated; please drop it.
- **Env surface:** several new `LITDATA_*` knobs, no docs, no validation (`floor > cap`, zeros, negatives). `LITDATA_ASSUMED_REQUEST_RATE` is wired but not listed in the PR body.
- **Tests** lock in the floor-32 “low bandwidth” behavior and the 0.001s → 500-slot jump. They do not cover the 256 KiB–1 MiB gap, hedge/range timing, or semaphore refresh.
- **Changelog** not updated (this repo usually does).

### Suggested direction

1. Keep static defaults until you have **enough observations of the right class** (N large GETs for bps, N small GETs for RTT), not “any 5 completions”.
2. For small objects, estimate RTT as `max(epsilon, duration - size/bps_ema)` once bps exists; do not feed full GET time into Little’s law.
3. Decide explicitly whether the **floor of 32** and **8 × workers** still apply when measured bps is low. If #870 is the goal, the floor has to move or the 429 case is unchanged.
4. Measure **one** stream (single range or unhedged GET), not parallel/hedged wall clocks.
5. Log when the budget changes (you already log on semaphore create) with `bps_ema`, `lat_ema`, `samples`, `budget` so the next bench is interpretable.

Please revise and add a bench that reports concurrency/permits and a slow-link or rate-limited case, not only Studio L4 + S3 val.

